PR #1, opened 2024-05-31
PB-554 Bootstrap the project

schtibe:
I think I'm going to split my work into multiple consecutive PRs so they can be reviewed chunk by chunk. I think at this point, everything might be subject to change quite often and fast, so we'll see how useful this is?

- Install Django and some other basic packages
- Setup docker-compose for the local DB
- Connect django to the DB
- Add some Makefile stuff from service-control

Maybe I do things a bit different than service-stac here and there, try to introduce new ideas and techs, but maybe I'll also discover that they're not a good idea down the road :)


docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+services:
+  db:
+    image: postgres:16
```
boecklic, line 3: Not sure if we get away without any geo-stuff in this service (which would require a postgis image), but that easily changeable so we can leave it like this for the moment.

Makefile
```diff
@@ -0,0 +1,65 @@
...
+
+# Commands
+PIPENV_RUN := pipenv run
+PYTHON := $(PIPENV_RUN) python3
+YAPF := $(PIPENV_RUN) yapf
+ISORT := $(PIPENV_RUN) isort
+
+
+.PHONY: setup
+setup: $(SETTINGS_TIMESTAMP) setup-s3-and-db setup-logs ## Setup the project by installing packages and activating venv
```
boecklic, line 33: `setup-s3-and-db` and `setup-logs` don't exist (yet)

boecklic (approved):
:+1: thanks. 

Have you other ideas regarding the `settings.py` handling compared to how we solved it for service stac (i.e. not having a `settings.py` in git but create it (and import from other `settings_*.py` files depending on environment)? Otherwise I'd rename it to `settings_dev.py` or sth (similar to stac)

docker-compose.yml
schtibe: OK, as long as the data is compatible I guess?

schtibe:
It's done in a later commit / PR.

PR #3, opened 2024-05-31
Pb-554-bootstrap-models

schtibe:
Add one basic model and expose it in admin


boecklic (approved):
as discussed: rename app `controller` to e.g. `provider`, add another for `dataset` etc...

PR #7, opened 2024-07-15
Feat pb coverage

schtibe:
Experimenting with coverage reports here


schtibe:
Closing for now as this isn't moving to a current priority.

PR #15, opened 2024-08-27
PB-897 Deployment work

schtibe:
Changing various stuff that's needed for service-control to run in k8s
* Missing / Wrong environment configurations
* Missing packages
* Missing wsgi file

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,10 +4,14 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-django = "~=5.0"
+django = "~=5.0.7"
 django-ninja = "~=1.1"
 psycopg2-binary = "~=2.9.5"
 django-environ = "~=0.11"
+gunicorn = "~=23.0"
+pyyaml = "~=6.0.2"
+gevent = "~=24.2"
+logging-utilities = "~=4.4.1"
 
 [dev-packages]
 yapf = "*"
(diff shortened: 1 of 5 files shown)


app/wsgi.py
```diff
@@ -1,16 +1,75 @@
...
+    import gevent.monkey
+    gevent.monkey.patch_all()
+"""
+WSGI config for project project.
 
 It exposes the WSGI callable as a module-level variable named ``application``.
 
 For more information on this file, see
-https://docs.djangoproject.com/en/5.0/howto/deployment/wsgi/
+https://docs.djangoproject.com/en/3.1/howto/deployment/wsgi/
```
ltshb, line 28: don't you want to point to django 5 here ?

ltshb (approved):
Good just a few minor comments

Pipfile
```diff
@@ -4,10 +4,14 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-django = "~=5.0"
+django = "==5.0.7"
```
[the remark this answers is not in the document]
schtibe: Indeed, thanks

app/wsgi.py
schtibe: Oh sure, I overwrote the default file with our custom file, and thus copy pasted the ancient link.

PR #21, opened 2024-09-12
Feat pb 902 setup models v1 in service control

asteiner-swisstopo:
Summary of changes:

1. Add three central Django models as a baseline to develop the API endpoints.
2. Clarify the setup instructions.
3. Add debugging support.

The three Django models are:

1. Provider: The legal entity that provides data for publication.
   For example: "Federal Office for the Environment" (FOEN)
2. Attribution: Information about the authors, displayed on the bottom left in the map viewer.
3. Dataset: A collection of data

`Attribution` and `Dataset` are in their own Django app `distributions` which was added with

```bash
python app/manage.py startapp distributions
```

The data model is inspired by Confluence page [Business Entity Model](https://ltwiki.adr.admin.ch:8443/display/PB/Business+Entity+Model) and [the API specs](https://2zvidp4qyh.apidog.io/).

ℹ️ Changes to Makefile targets are mostly copy-paste from service-stac:

- `make setup`: Instructions after activation of a venv are only executed once the subshell is exitted again,
  so move them to before `pipenv shell`.
- `make serve-debug`: Facilitate attaching a debugger to a running server.
- `make test-debug`: Facilitate attaching a debugger to a running unit test.
  Although there are currently no static files, we follow the example of `service-stac`
  and collect the static files into one folder to avoid future warnings.


app/distributions/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+    _context = "Attribution Model"
+
+    def __str__(self) -> str:
+        return str(self.name_en)
+
+    name_de = models.CharField(_(_context, "Name (German)"))
+    name_fr = models.CharField(_(_context, "Name (French)"))
+    name_en = models.CharField(_(_context, "Name (English)"))
+    name_it = models.CharField(_(_context, "Name (Italian)"), blank=True)
```
boecklic, line 15: I think we should allow the non-required fields to be empty on db-level as well, and only set fallback value on the fly when requested. Otherwise it will be difficult to track which values where actually provided in API calls.

app/distributions/tests/test_attribution_model.py
```diff
@@ -0,0 +1,101 @@
...
+        self.assertEqual(attribution.name_rm, name_rm)
+
+        self.assertEqual(attribution.description_de, description_de)
+        self.assertEqual(attribution.description_fr, description_fr)
+        self.assertEqual(attribution.description_en, description_en)
+        self.assertEqual(attribution.description_it, description_it)
+        self.assertEqual(attribution.description_rm, description_rm)
+
+        self.assertEqual(attribution.provider.acronym_de, "ENSI")
+
```
boecklic, line 56: I'd maybe add another test case for a model with just the required values

boecklic (commented):
Nice, thanks!

app/distributions/models.py
asteiner-swisstopo: Not sure I understand what you mean.

The `blank=True` is only taken into account for form validation, not on what is written into the DB. You can see that for example in the admin page: The `blank=True` fields are shown in regular font, not bold as the required ones. So one can still write null or empty string (blank) to the DB if it's not via a form.

Probably I should also add the comment from [provider/models.py](https://github.com/geoadmin/service-control/blob/feat-PB-902-setup-models-v1-in-service-control/app/provider/models.py#L12) here:
```py
''''
Note: The "blank=False" for a model field doesn't prevent DB changes.
          It only has an effect on form validation.
''''
```

My impression is that this `blank=True` only creates confusion without a big gain (form validiation). So I propose to remove it.

Ok with you?

app/distributions/tests/test_attribution_model.py
asteiner-swisstopo: Good point, thanks. Will do as we are clear on what to do with the required vs optional fields. Same for `ProviderTestCase`.

schtibe:
@asteiner-swisstopo I just noticed that you merged `develop` into this branch here. AFAIK this could lead to a messy git history which [we want to avoid](https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#goal). Maybe rebase your branch on `develop` and remove the merge commit (I can help you with that if you don't know how)

asteiner-swisstopo:
> @asteiner-swisstopo I just noticed that you merged `develop` into this branch here. AFAIK this could lead to a messy git history which [we want to avoid](https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#goal). Maybe rebase your branch on `develop` and remove the merge commit (I can help you with that if you don't know how)

Good point thanks, I had never done that before.

I changed the git history as follows:
1. `git fetch` to update the local clone
2. `git log develop..feat-PB-902-setup-models-v1-in-service-control` to identify the hash of the commit _before_ the merge commit (`015124505d08575c1791340beec78fa2cdd3ed83`).
3. `git rebase -i 015124505d08575c1791340beec78fa2cdd3ed83^` and, in the interactive mode (text editor), delete the line with the commit that came in with the merge ("PB-897 Fix the year in the license").
4. `git push --force` to sync the changes

To me it looks now as expected. Please let me know if there would have been an easier way to do it.

app/distributions/models.py
boecklic: Clarified in personal communication: `blank=True` allows empty fields in a form, `null=True` allows `null` values in the database. Often it's either `blank=True,null=True` or `blank=False,null=False`, but certain use cases can require only one of both `True`

asteiner-swisstopo: Should be adapted now, could you please have another look?

app/distributions/tests/test_attribution_model.py
asteiner-swisstopo: Added new test cases for both:
- `test_object_created_in_db_with_optional_fields_null`: A model just with the required values
- `test_raises_exception_when_creating_db_object_with_mandatory_field_null`: To illustrate what happens if you set a required field to `NULL` / `None`.

.gitignore
```diff
@@ -1,2 +1,9 @@
-.env
+# Byte-compiled / optimized / DLL files
 __pycache__
+
+# Distribution / packaging
+.env
+var/
+
+# IDE config files
+.vscode
```
schtibe, line 9: 👍

README.md
```diff
@@ -84,11 +84,17 @@ in it:
...
-the debugger is attached, which can most easily done by hitting f5.
+Alternatively, create the file via menu "Run" > "Add Configuration" by choosing
+
+- Debugger: Python Debugger
+- Debug Configration: Remote Attach
+- Hostname: `localhost`
+- Port number: `5678`
+
+Now you can start the server with `make serve-debug`.
+The bootup will wait with the execution until the debugger is attached, which can most easily done by hitting F5.
```
schtibe, line 97: Very nice!

app/config/settings_base.py
```diff
@@ -121,6 +123,7 @@
 # https://docs.djangoproject.com/en/5.0/howto/static-files/
 
 STATIC_URL = 'static/'
+STATIC_ROOT = BASE_DIR / 'var' / 'www' / 'service_control' / 'static_files'
```
schtibe, line 126: This string seems a bit weird

schtibe (approved):
Very nice!

schtibe:
> > @asteiner-swisstopo I just noticed that you merged `develop` into this branch here. AFAIK this could lead to a messy git history which [we want to avoid](https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#goal). Maybe rebase your branch on `develop` and remove the merge commit (I can help you with that if you don't know how)
> 
> Good point thanks, I had never done that before.
> 
> I changed the git history as follows:
> 
>     1. `git fetch` to update the local clone
> 
>     2. `git log develop..feat-PB-902-setup-models-v1-in-service-control` to identify the hash of the commit _before_ the merge commit (`015124505d08575c1791340beec78fa2cdd3ed83`).
> 
>     3. `git rebase -i 015124505d08575c1791340beec78fa2cdd3ed83^` and, in the interactive mode (text editor), delete the line with the commit that came in with the merge ("[PB-897](https://jira.swisstopo.ch/browse/PB-897) Fix the year in the license").
> 
>     4. `git push --force` to sync the changes
> 
> 
> To me it looks now as expected. Please let me know if there would have been an easier way to do it.

Cool! No I would have done more or less the same!

app/provider/tests/test_provider_model.py
```diff
@@ -1,22 +1,88 @@
...
+        name_fr = "Office fédéral de l'environnement"
+        name_en = "Federal Office for the Environment"
+        name_it = "Ufficio federale dell'ambiente"
+        name_rm = "Uffizi federal per l'ambient"
+        acronym_de = "BAFU"
+        acronym_fr = "OFEV"
+        acronym_en = "FOEN"
+        acronym_it = "UFAM"
+        acronym_rm = "UFAM"
+        Provider.objects.create(
```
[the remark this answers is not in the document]
asteiner-swisstopo: Great idea, thanks. Will adapt :+1:

app/distributions/models.py
```diff
@@ -0,0 +1,43 @@
...
+    class Meta:
+        verbose_name = _("dataset")
+        verbose_name_plural = _("datasets")
+
+    def __str__(self) -> str:
+        return str(self.slug)
+
+    slug = models.SlugField(_("Slug"))
+    created = models.DateTimeField(_("Created"), auto_now_add=True)
+    updated = models.DateTimeField(_("Updated"), auto_now=True)
```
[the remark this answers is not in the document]
asteiner-swisstopo: Thanks for the hint. I am going to leave it as it is but let's keep this limitation in mind.

app/config/settings_base.py
asteiner-swisstopo: Totally. I took this from [settings_prod.py in service-stac](https://github.com/geoadmin/service-stac/blob/674b1ab1212a79dd7b8463df642300ec03b535cf/app/config/settings_prod.py#L180). What would be a better path?

benschs (approved):

app/provider/tests/test_provider_model.py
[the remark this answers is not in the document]
asteiner-swisstopo: :heavy_check_mark: Changed according to your suggestion.

app/config/settings_base.py
boecklic: `BASE_DIR` is a `Path` object from `pathlib`, not a raw string. `pathlib` supports or even suggests to construct paths like this. So I think it's ok to leave it like this. If you want you can add a comment to make this clear.

boecklic (approved):

PR #22, opened 2024-09-23
PB-903 Configure pytest

asteiner-swisstopo:
Summary of changes:
1. Make it possible to run the test just by typing `pytest` in the console (or, as before, `make test`). We already had Python module `django-pytest` installed but it was not used; we had to go via `manage.py test` to run tests. Pytest is faster and its output is easier to read.
2. Remove Make target `test-debug`. I see no point in this, it was copied from `service-stac`. Both the pytest CLI and the IDE can handle debugging without this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,11 +148,6 @@ start-local-db: ## Run the local db as docker container
 .PHONY: test
 test: ## Run tests locally
 	$(TEST)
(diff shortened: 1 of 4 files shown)


benschs (approved):

boecklic (approved):
:+1:

PR #24, opened 2024-09-25
PB-903: Add endpoint getProvider

asteiner-swisstopo:
This implements end point `getProvider` as defined [in the API specs](https://2zvidp4qyh.apidog.io/api-8832236).

Example response:

```json
{
    "id": "1",
    "name": "Bundesamt für Umwelt",
    "name_translations": {
        "de": "Bundesamt für Umwelt",
        "fr": "Office fédéral de l'environnement",
        "en": "Federal Office for the Environment",
        "it": "Ufficio federale dell'ambiente",
        "rm": "Uffizi federal per l'ambient",
    },
    "acronym": "BAFU",
    "acronym_translations": {
        "de": "BAFU",
        "fr": "OFEV",
        "en": "FOEN",
        "it": "UFAM",
        "rm": "UFAM",
    }
}
```

On the endpoints;
   1. The endpoint takes the language into account: Either via the query parameter (`providers/{provider_id}/?lang="de"`) or via the HTTP header "[Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)". The query parameter takes precedence, otherwise the language with the largest q-factor in the header goes first (e.g. "q=0.8" > "q=0.6"). If none is specified, we fall back to English. This affects fields "name" and "acronym".
   2. To parse the values of "Accept-Language" (like `"ru;q=0.9, de;q=0.7, *, rm;q=0.8"`), Python package [acceptlang](https://github.com/piotrszyma/acceptlang) is taken. It does not take _that_ much of an effort to do this yourself, so if this introduces unnecessary dependencies, I am happy to remove it.
    :point_up: EDIT: Removed the dependency. Instead, just take the first known language, ignoring q factors.
   3. A dedicated function was created to extract a field with a specific suffix from a Django model. For example, if the API query set "de" as language, you would get `Provider.name_de` and `Provider.acronym_de`. I extracted it to a function for reuse later when have to do something similar with `Attribution.name_de` and `Attribution.description_de` etc.
       :warning: I find this quite ugly but, if I understand correctly, Django models have to be a flat list. So I could not do something like `Provider.name["de"]`. If you see a better solution, please let me know.
       :information_source: This lives in its own subpackage `utils`, same for the header parsing. Not sure if this is the best place but looks good enough for me to start with.

Django/Ninja specifics:
   1. Split the API into routers in order to be able to have multiple files defining the endpoints. See the [Django Ninja docs](https://django-ninja.dev/guides/routers/) for more.
   2. If a field in the response is not defined, it is skipped. This can be conveniently done with option [exlude_none](https://django-ninja.dev/reference/operations-parameters/#exclude_none).
   3. The `id` is equivalent to the primary key of the corresponding record in the DB. Probably not the best solution in terms of security (could guess other entries) and maybe usability, but good enough for a start.
      :warning: Testing against primary keys is not ideal as it makes you dependent on having your test run in isolation. If you accept the performance penalty, it can be achieved with [TransactionTestCase.reset_sequences](https://docs.djangoproject.com/en/5.1/topics/testing/advanced/#django.test.TransactionTestCase.reset_sequences).

We decided to tackle the other end points in separate PRs. The endpoints are all quite similar, so the learnings from this one can be taken into account for the others from the start.


boecklic (commented):
:+1:

app/provider/api.py
```diff
@@ -0,0 +1,89 @@
...
+          is taken (Example: "de;q=0.7, rm;q=0.8" --> "rm")
+        - Subtags in the header are ignored. So "en-US" is interpreted as "en".
+        - Wildcards ("*") are ignored.
+    """
+    if not lang:
+        if "Accept-Language" in request.headers:
+            lang_header = request.headers["Accept-Language"]
+            lang = extract_lang(lang_header)
+        else:
+            lang = LanguageCode.ENGLISH
```
benschs, line 67: Maybe move all of this into the `extract_lang(request)` function. Reduces overhead in the api functions, as well as have a single place that defines the default language.

app/config/api.py
```diff
@@ -0,0 +1,11 @@
...
+
+api = NinjaAPI()
+
+
+@api.get('/checker')
+def checker(request):
+    return {"success": True, "message": " OK"}
+
+
+api.add_router("/provider", "provider.api.router")
```
benschs, line 11: I guess the path should be `/providers` or `/api/providers`. Not sure where to best add the `/api` prefix, as it will be used by all apps.

app/provider/api.py
```diff
@@ -0,0 +1,89 @@
...
+from django.shortcuts import get_object_or_404
+
+from .models import Provider
+from .schemas import ProviderSchema
+from .schemas import TranslationsSchema
+
+router = Router()
+
+
+@router.get("/providers/{provider_id}", response={200: ProviderSchema}, exclude_none=True)
```
benschs, line 16: As we add the `/providers` prefix in `config/api.py` we don't need it here.

app/provider/api.py
```diff
@@ -0,0 +1,89 @@
...
+        - Wildcards ("*") are ignored.
+    """
+    if not lang:
+        if "Accept-Language" in request.headers:
+            lang_header = request.headers["Accept-Language"]
+            lang = extract_lang(lang_header)
+        else:
+            lang = LanguageCode.ENGLISH
+
+    provider_object = get_object_or_404(Provider, id=provider_id)
```
benschs, line 69: Do we need some validation that `provider_id` is an integer?

app/config/api.py
benschs: I think it would also be better to import the router and add it as an object as it will make any refactoring/renaming easier in the future.

app/provider/api.py
asteiner-swisstopo: You are right. As discussed, extracted it to a dedicated function such that there is only one place where the default language is defined.

app/provider/api.py
asteiner-swisstopo: Well spotted, should be `int`. Ninja would then validate the query parameter type.

app/utils/translation.py
```diff
@@ -0,0 +1,48 @@
...
+    Two-letter language codes
+    """
+    GERMAN = "de"
+    FRENCH = "fr"
+    ITALIAN = "it"
+    ROMANSH = "rm"
+    ENGLISH = "en"
+
+
+def get_translation(
```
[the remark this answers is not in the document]
asteiner-swisstopo: Ok thanks, never heard of JSONB fields. Will have a look with them :+1:

asteiner-swisstopo (commented):
Thanks for the great review @boecklic and @benschs! I have addressed your comments except the one on JSONB which is not yet clear to me.

Other than that feel free to have another look if my changes correspond to your suggestions.

boecklic (approved):
👍🏻

app/utils/translation.py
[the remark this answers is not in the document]
asteiner-swisstopo: For the record, we decided to have a look at this in a subsequent iteration. Good enough for the moment.

PR #25, opened 2024-10-02
PB-903 Endpoints for Attribution and Dataset

asteiner-swisstopo:
This is the follow-up PR to [the addition of endpoint `api/providers/{provider_id}`](https://github.com/geoadmin/service-control/pull/24).

Summary of changes:
1. 5 new GET endpoints:
   1. `api/providers`
   2.  `api/attributions/{attribution_id}`
   3. `api/attributions`
   4. `api/datasets/{datasets_id}`
   5. `api/datasets`
6. Minor refactoring in `app/utils/language.py` to refer to the same constant, so we have only one place to change if we want to change the default language.

Example responses:

For `api/providers` ([specs](https://2zvidp4qyh.apidog.io/api-8828563))

```json
{
    "items": [
        {
            "id": 1,
            "name": "Office fédéral de l'environnement",
            "name_translations": {
                "de": "Bundesamt für Umwelt",
                "fr": "Office fédéral de l'environnement",
                "en": "Federal Office for the Environment",
                "it": "Ufficio federale dell'ambiente",
                "rm": "Uffizi federal per l'ambient",
            },
            "acronym": "OFEV",
            "acronym_translations": {
                "de": "BAFU",
                "fr": "OFEV",
                "en": "FOEN",
                "it": "UFAM",
                "rm": "UFAM",
            }
        }
    ]
}
```

For `api/attributions/123` ([specs](https://2zvidp4qyh.apidog.io/api-9150426)):

```json
{
    "id": 123,
    "name": "FOEN",
    "name_translations": {
        "de": "BAFU",
        "fr": "OFEV",
        "en": "FOEN",
        "it": "UFAM",
        "rm": "UFAM",
    },
    "description": "Federal Office for the Environment",
    "description_translations": {
        "de": "Bundesamt für Umwelt",
        "fr": "Office fédéral de l'environnement",
        "en": "Federal Office for the Environment",
        "it": "Ufficio federale dell'ambiente",
        "rm": "Uffizi federal per l'ambient",
    },
    "provider_id": 1,
}
```

For `api/attributions` ([specs](https://2zvidp4qyh.apidog.io/api-9150424)):

```json
{
    "items": [
        {
            "id": 1,
            "name": "OFEV",
            "name_translations": {
                "de": "BAFU",
                "fr": "OFEV",
                "en": "FOEN",
                "it": "UFAM",
                "rm": "UFAM",
            },
            "description": "Office fédéral de l'environnement",
            "description_translations": {
                "de": "Bundesamt für Umwelt",
                "fr": "Office fédéral de l'environnement",
                "en": "Federal Office for the Environment",
                "it": "Ufficio federale dell'ambiente",
                "rm": "Uffizi federal per l'ambient",
            },
            "provider_id": 2,
        }
    ]
}
```

For `api/datasets/123` ([specs](https://2zvidp4qyh.apidog.io/api-9156043)):

```json
{
    "id": 123,
    "slug": "ch.bafu.neophyten-haargurke",
    "created": "2023-09-26T12:39:29.964574Z",
    "updated": "2023-09-26T12:39:29.964574Z",
    "provider_id": 1,
    "attribution_id": 2,
}
```

For `api/datasets` ([specs](https://2zvidp4qyh.apidog.io/api-9156026)):

```json
{
    "items": [{
        "id": 123,
        "slug": "ch.bafu.neophyten-haargurke",
        "created": "2023-09-26T12:39:29.964574Z",
        "updated": "2023-09-26T12:39:29.964574Z",
        "provider_id": 1,
        "attribution_id": 2,
    }]
}
```


app/provider/api.py
```diff
@@ -7,12 +7,39 @@
...
 
 from .models import Provider
+from .schemas import ProviderListSchema
 from .schemas import ProviderSchema
 from .schemas import TranslationsSchema
 
 router = Router()
 
 
+def to_response(model: Provider, lang: LanguageCode) -> ProviderSchema:
```
msom, line 17: might name it `provider_to_response` for consistency

msom (approved):
Looks good,  only some optional remarks. I like your clean code and good test coverage!

app/distributions/api.py
```diff
@@ -0,0 +1,149 @@
...
+    Get all attributions, return translatable fields in the given language.
+
+    For more details on how individual attributions are returned, see the
+    corresponding endpoint for a specific attribution.
+    """
+    models = Attribution.objects.order_by("id").all()
+    lang_to_use = get_language(lang, request.headers)
+
+    responses = [attribution_to_response(model, lang_to_use) for model in models]
+    return AttributionListSchema(items=responses)
```
schtibe, line 112: Is it necessary to explicitly call the Schema-Serializer? From the [docs](https://django-ninja.dev/tutorial/other/crud/#list-of-objects) it seems that you could only return the list and let Ninja handling the Serialization?

app/distributions/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+    id: str
+    name: str
+    name_translations: TranslationsSchema
+    description: str
+    description_translations: TranslationsSchema
+    provider_id: str
+
+
+class AttributionListSchema(Schema):
+    items: list[AttributionSchema]
```
schtibe, line 17: I wonder if this could directly be a list instead of being nested in items (but it's done like specified in apidog) cc @boecklic

app/distributions/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+    id: str
+    slug: str
+    created: datetime
+    updated: datetime
+    provider_id: str
+    attribution_id: str
+
+
+class DatasetListSchema(Schema):
+    items: list[DatasetSchema]
```
schtibe, line 30: See comment above

app/distributions/api.py
```diff
@@ -0,0 +1,149 @@
...
+            en=model.description_en,
+            it=model.description_it,
+            rm=model.description_rm,
+        ),
+        provider_id=str(model.provider.id),
+    )
+    return response
+
+
+@router.get("attributions/{attribution_id}", response={200: AttributionSchema}, exclude_none=True)
```
schtibe, line 47: Doc says, `exclude_none=True` leads to exclusion of None fields. Does that mean that for instance `name_translations.it` would be omitted if empty? Does that even have an effect when we serialize manually (see comment below).
(Maybe I just don't understand Ninja well enough yet 😀)

app/schemas.py
```diff
@@ -0,0 +1,9 @@
+from ninja import Schema
+
+
+class TranslationsSchema(Schema):
```
schtibe, line 4: Maybe a better name for this could be something that indicates a Translated String? Since this is basically a string multiplied by N languages

app/distributions/api.py
schtibe: Ok I just saw from the tests that that's apparently what we want 😅

app/distributions/tests/test_api.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def test_get_datasets_returns_single_dataset_as_expected(self):
+
+        provider = Provider.objects.create(acronym_de="BAFU")
+        attribution = Attribution.objects.create(
+            name_de="Kantone",
+            provider=provider,
+        )
+        model_fields = {
+            "slug": "ch.bafu.neophyten-haargurke",
```
schtibe, line 689: 😬

schtibe (approved):
Very nice, I just a have some clarification remarks, otherwise, good job!

app/distributions/tests/test_api.py
```diff
@@ -0,0 +1,764 @@
...
+            "name_it": "UFAM",
+            "name_rm": "UFAM",
+            "description_de": "Bundesamt für Umwelt",
+            "description_fr": "Office fédéral de l'environnement",
+            "description_en": "Federal Office for the Environment",
+            "description_it": "Ufficio federale dell'ambiente",
+            "description_rm": "Uffizi federal per l'ambient",
+            "provider": provider,
+        }
+        Attribution.objects.create(**model_fields)
```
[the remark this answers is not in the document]
asteiner-swisstopo: > Why not just `model = Attribution.objects.create(**model_fields)`?

Totally, that would be easier. Will fix.

> Also you might consider adding this to a setup function to avoid duplication of code (and also split test cases to endpoints)

Good point, I had also thought about it. I am not a fan of fixtures because I cannot see at a glance what the test does. Sometimes it's still a benefit when the setup is very long. So let me have another look, if it's duplicated often I can adapt it.

app/provider/api.py
asteiner-swisstopo: Good point, will fix.

app/distributions/api.py
asteiner-swisstopo: Good point, I originally wanted to do that. The thing is that we have it wrapped into the `items=[...]` keyword, so I need a Schema that spells this out.

Do you know what's the benefit of this `items`? It's in every example of [the API specs](https://2zvidp4qyh.apidog.io/api-9150424) but I am not sure what is its purpose.

app/distributions/api.py
```diff
@@ -0,0 +1,149 @@
...
+from .schemas import DatasetSchema
+
+router = Router()
+
+
+def attribution_to_response(model: Attribution, lang: LanguageCode) -> AttributionSchema:
+    """
+    Transforms the given model using the given language into a response object.
+    """
+    response = AttributionSchema(
```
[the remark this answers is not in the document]
asteiner-swisstopo: Not sure what you mean, this is not returning a list. This is to make sure that some special fields that depend on the language are filled with the correct value (here: "name" and "description"). We cannot just map fields from the model 1-1 on a schema.

Does that answer your question?

app/distributions/schemas.py
asteiner-swisstopo: I wondered the same, assumed it would serve a certain end as it's used in every example. :sweat_smile:

app/schemas.py
asteiner-swisstopo: :thinking: You are right to say that "translation" is a simplification, after all a "translation" could not only contain the translated text itself but also notes, a date, the translator's name etc. I find that in the given context, it is an acceptable simplification to assume that
- translated string
- == translated text
- == translation

I don't like using "string" in the name here, too implementation-specific. If you have another good suggestion let me know, otherwise I would propose to stick with "translations" for brevity.

app/distributions/tests/test_api.py
asteiner-swisstopo: 🦠

app/schemas.py
schtibe: OK for me. I was rather thinking in the direction of it could be a TranslatedList or TranslatedAnything but maybe that's too far fetched.

app/distributions/api.py
schtibe: I guess the way would be to return a dict like so

```
return {"items": responses}
```

Since you specify `AttributionListSchema` in the function decorator my assumption is that Ninja takes care of actually schema-ify the data afterwards.

app/distributions/api.py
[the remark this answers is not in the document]
schtibe: I sort of mean the same as with my comment on the attributions GET endpoint (l 112).

The endpoint calls `attribution_to_response`  which converts the data explicitly into a `AttributionSchema`.
It returns it to the endpoint (`attribution()`) , which then returns it to Ninja.

When my assumption is correct, `attribution_to_response()` could only return a dict, which is then (through going to `attribution()`)  passed to Ninja, which could do the serialization.

boecklic (approved):
Nice!

app/distributions/api.py
asteiner-swisstopo: Ok, so I would suggest to keep the schema because
1. We might want to add more fields than just "items".
2. It validates that we give it a list. Otherwise we might give it some other other iterable by accident, like a set.

Concerns with that @schtibe?

schtibe:
Sorry, something that I've missed until now: it seems to be common practisce to use version numbers in the API URL, such as `/api/v0.1/providers`

app/distributions/api.py
schtibe: Not sure if we're still talking about the same thing 😅 I made a PR in your PR to demonstrate what I mean:
https://github.com/geoadmin/service-control/pull/27

schtibe (changes requested):

app/provider/api.py
```diff
@@ -7,12 +7,39 @@
...
 
 router = Router()
 
 
+def provider_to_response(model: Provider, lang: LanguageCode) -> ProviderSchema:
+    """
+    Transforms the given model using the given language into a response object.
+    """
+    response = ProviderSchema(
+        id=str(model.id),
```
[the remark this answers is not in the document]
asteiner-swisstopo: Well spotted!

Yes, the reason is that the numerical ID is something easy to start with (ID == private key of the corresponding record in the DB) but later probably we should have something else. The numerical ID has a security issue (you can guess other IDs) and it is not very self-describing. So either a cryptic UID or something even smarted would be a way to go.

I'll add a comment to the schema to make the point clear.

app/distributions/api.py
asteiner-swisstopo: Ok, discussed this in person and left a [conclusion in the corresponding PR](https://github.com/geoadmin/service-control/pull/27#issuecomment-2399904098). I apply the relevant changes here.

app/provider/api.py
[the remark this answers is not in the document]
asteiner-swisstopo: For the record, we agreed that setting those ID fields to `int` would more properly reflect the current state. We can still change it later to `str`.

schtibe (approved):

PR #26, opened 2024-10-04
PB-897 Fix the checker URL

schtibe:
The checker URL wasn't reachable anymore after changes to the URL structure


app/config/api.py
```diff
@@ -1,11 +1,10 @@
 from ninja import NinjaAPI
 from provider.api import router as provider_router
 
-root = NinjaAPI()
 api = NinjaAPI()
 
 
-@root.get('/checker')
+@api.get('/checker')
```
asteiner-swisstopo, line 7: :thinking: We register the `api` object in `urls.py` as

```py
urlpatterns = [
    path('api/', api.urls),
    path('admin/', admin.site.urls),
]
```

Doesn't that mean, that with this change the checker is reachable at `api/checker` instead of `/checker`?

asteiner-swisstopo: I see three reasons for putting it under root instead of `/api`:
1. In service-stac, we have it under root [if I understand correctly](https://github.com/geoadmin/service-stac/blob/8566c53d0300013a2ffe2de8011229d26ec616c0/app/config/urls.py#L33) (`/checker`)
2. For [Kubernetes pods](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), usually it is under root (`/healthz`).
3. As a beginner, I would intuitively expect stuff to interact with under `/api`. Infrastructure-related endpoints would be somewhere else, for example under root.

But no strong opinion. If it's too much of a hassle to change or if you prefer `/api/checker`, I am ok with keeping it.

What is your preference?

schtibe: I agree! I'm gonna change it ASAP

app/config/urls.py
```diff
@@ -17,9 +17,9 @@
 from django.contrib import admin
 from django.urls import path
 
-from .api import api
+from .api import root
 
 urlpatterns = [
-    path('api/', api.urls),
+    path('api/', root.urls),
```
asteiner-swisstopo, line 23: :thinking: Not sure this is correct.

As I understand it, now we have only `/api/checker` registered. But instead we want
- Endpoint for the users: `/api/providers`, `/api/datasets` etc
- Endpoint for infra: `/checker`

So shouldn't it be something like this instead?

```py
urlpatterns = [
    path('api/', api.urls),
    path('/', root.urls),
    path('admin/', admin.site.urls),
]
```

schtibe: Thanks, messed it up. Changed it to not use NinjaAPI for the checker as discussed in person.

app/config/test_checker.py
```diff
@@ -0,0 +1,21 @@
+from django.test import Client
+from django.test import TestCase
+
+
+class CheckerUrlTestCase(TestCase):
```
asteiner-swisstopo, line 5: Does not need `django.test` here as no interaction with the DB is necessary.

So just this would be enough:
```py
from django.test import Client


def test_checker_url():
    #....
```
Any concerns with that?

app/config/test_checker.py
```diff
@@ -0,0 +1,21 @@
+from django.test import Client
```
asteiner-swisstopo, line 1: Two concerns about the file:
1. Shouldn't it be in `tests` folder like all the others?
2. Should it be `test_urls` to reflect that it checks stuff in `urls.py`? On the other hand, we also have `test_attribution_model.y` even though it then should be `test_models.py`. So fine with both, just wanted to raise the point because I am used to have one test file per module (like `mymodule.py` --> `test_mymodule.py`)

app/config/test_checker.py
schtibe: I thought that inheriting from TestCase is necessary :thinking:  Mabye there would be a better way without the Django baggage?

app/config/test_checker.py
schtibe: I intentionally didn't add a folder as I don't expect to have more tests in here, since this is the `config` folder after all.

app/config/test_checker.py
schtibe: Changed it to use `unittest.TestCase` directly :+1:

schtibe:
@asteiner-swisstopo I reverted back to the django TestCase, see the commit on why.

app/config/test_checker.py
schtibe: I moved it to a folder. Why not. Keep the same pattern across all apps :+1:

asteiner-swisstopo (approved):
All good, thanks :white_check_mark:

Regarding the pytest, we might have been able to solve it according to [the pytest docs](https://pytest-django.readthedocs.io/en/latest/database.html) with a decorator

```py
@pytest.mark.django_db
```

but not even convinced that this is better.

PR #27, opened 2024-10-07
PB-903 Drop the explicit schema invocations

schtibe:
According to the docs, it isn't necessary to invoke the Schemas by ourselves. The endpoint functions return the data as lists, querysets or dicts and the Framework handles the conversion to the schema. I don't see an immediate drawback to our version, but I'm also not so comfortable to deviate from the standard way proposed by the docs.

From the [docs](https://django-ninja.dev/guides/response/), NinjaAPI seems to do quite some tasks with the returned data itself and in my opinion we shouldn't be doing this ourselves unless we have a good reason to.
Also, there might be performance optimizations being done by NinjaAPI (consider especially when returning entire querysets)

### DIFF
--- a/app/distributions/api.py
+++ b/app/distributions/api.py
@@ -1,5 +1,4 @@
 from ninja import Router
-from schemas import TranslationsSchema
 from utils.language import LanguageCode
 from utils.language import get_language
 from utils.language import get_translation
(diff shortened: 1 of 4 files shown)


app/distributions/api.py
```diff
@@ -17,30 +16,30 @@
...
-            it=model.description_it,
-            rm=model.description_rm,
-        ),
-        provider_id=str(model.provider.id),
-    )
+    response = {
+        "id": str(model.id),
+        "name": get_translation(model, "name", lang),
+        "name_translations": {
+            "de": model.name_de,
```
schtibe, line 27: As a demonstration that Ninja handles the transformation, remove the field "de" here and Ninja throws a `ValidationError`

app/distributions/api.py
```diff
@@ -132,8 +117,7 @@ def dataset(request: HttpRequest, dataset_id: int):
     Get the dataset with the given ID.
     """
     model = get_object_or_404(Dataset, id=dataset_id)
-    response = dataset_to_response(model)
-    return response
+    return model
```
schtibe, line 120: There's no need to explicitly call the DatasetSchema. Ninja does it when we tell it via decorator

app/distributions/api.py
asteiner-swisstopo: I understand but I don't see the advantage of using a dict over a schema here.

With a structured object like the schema, we have these advantages:
1. Clearer interface: It returns an object with a defined set of fields, not just a dict that could contain anything.
2. Facilitates type checking early: The type checker (mypy) can already find issues at build time instead of at runtime.
3. Easier coding: The IDE can suggest the field names and assist in renaming fields everywhere they are used.

What speaks in favor of using a dict instead?

app/distributions/api.py
asteiner-swisstopo: I see, nice! Didn't know it would be able to convert an integer to a string by itself.

app/distributions/api.py
```diff
@@ -109,21 +108,7 @@ def attributions(request: HttpRequest, lang: LanguageCode | None = None):
...
-    """
-    return DatasetSchema(
-        id=str(model.id),
-        slug=model.slug,
-        created=model.created,
-        updated=model.updated,
-        provider_id=str(model.provider.id),
-        attribution_id=str(model.attribution.id),
-    )
+    return {"items": responses}
```
asteiner-swisstopo, line 111: :thinking: Fine with me here because the decorator defines the return type and it makes it shorter.

Still, the same concerns as mentioned above apply:
1. Less explicit: I have to check the decorator to see what is returned ultimately.
2. No type checking: Not sure if mypy would alert me here if `items` was something different from a list.
3. Less coding assistance: Anything can go into that dict.

So fine if you want to use a dict here. But not a fan :wink:

asteiner-swisstopo:
For the record, we agreed to incorporate some changes of this in the target branch directly:
- Directly return the model or use a dict where it is convenient as a return type. Ninja then does the implicit conversion.
- Keep using the Schema in dedicated functions in order to make the intent clear.

@schtibe Can you please close/cancel this PR?

PR #28, opened 2024-10-09
Tighten Some Type Checking Rules

msom:
The type checker is more useful when more checks are activated, as it allow to to checking for undefined attributes or missing overloads, for example.

I suggest that we enable some additional checks:
- enforce type definitions on methods (`disallow_untyped_defs`, `disallow_incomplete_defs`, `disallow_untyped_decorators`)
- check types within untyped methods (`check_untyped_defs`)
- show additional warnings for unused imports, missing return statements and unreachable code (`warn_unused_ignores`, `warn_no_return`, `warn_unreachable`)
- restrict equality checks to same types (`strict_equality`)

But only for our own code (`ignore_missing_imports`) and not for tests (`exclude`).


asteiner-swisstopo:
General comment: Wouldn't it be easier to just set the strict mode and maybe disable things we absolutely don't want?

I had a go at this in branch `sandbox-PB-1056-mypy-in-strict-mode` (derived from this branch) and I had to adapt only very sensible things imho:
https://github.com/geoadmin/service-control/compare/feat-pb-1056-tighten-mypy..sandbox-PB-1056-mypy-in-strict-mode

Explicitly listing all the rules we want has the risk that mypy adds checks in the future and we don't add them here. Going with the `--strict` we are sure to always have them.

msom:
@asteiner-swisstopo it's now as strict as your branch. I didn't use the monkeypatch part, as this results in a linting error. I think it's better to use a `type:ignore` here.

mypy.ini
```diff
@@ -4,6 +4,22 @@ mypy_path = ./app,./app/stubs
 plugins =
     mypy_django_plugin.main
 explicit_package_bases = True
+exclude = ^.*[/\\]test_.*
+
+# add some additional checks, but not quite strict
+strict = True
+disallow_untyped_calls = False
```
schtibe, line 11: What's the reason for this rule?

schtibe (approved):
Very nice, just a small question!

mypy.ini
msom: > Disallows calling functions without type annotations from functions with type annotations.
Else we would need properly typed stubs for every not-typed library.

asteiner-swisstopo: I see that if we enable this, it only affects one place [in wsgi.py](https://github.com/geoadmin/service-control/blob/feat-pb-1056-tighten-mypy/app/wsgi.py#L78):
```py
StandaloneApplication(application, options).run()
```
because `Baseapplication.run()` is in the untyped package `gunicorn`.

There is the option to exclude untyped packages as described in [the docs for `disallow_untyped_calls` ](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_calls):
> If you want to selectively disable this check for all your code that imports some.library you should instead use [untyped_calls_exclude](https://mypy.readthedocs.io/en/stable/config_file.html#confval-untyped_calls_exclude), for example:
>
> ```toml
> [mypy]
> disallow_untyped_calls = True
> untyped_calls_exclude = some.library

So in our case:

```
untyped_calls_exclude = gunicorn
```

This is not going to resolve the case above because `StandaloneApplication` inherits from `gunicornBaseapplication`. Mypy is apparently too limited to take the inheritance hierarchy into account, see mypy issue ["--untyped-calls-exclude should consider inheritance"](https://github.com/python/mypy/issues/16651). So here, we would still need to a `# type: ignore`.

I would keep the rule because there must be good reasons that it's set as a default. But no strong opinion, to me it does not seem to be the most important one.

How do you see it given that there is `untyped_calls_exclude`?

mypy.ini
```diff
@@ -4,6 +4,23 @@ mypy_path = ./app,./app/stubs
...
+ignore_missing_imports = True
+implicit_reexport = true
```
asteiner-swisstopo, line 13: This is already included in strict mode, see [the mypy.ini equivalent to `--strict`](https://mypy.readthedocs.io/en/stable/existing_code.html#getting-to-strict). Can be removed here.

mypy.ini
```diff
@@ -4,6 +4,23 @@ mypy_path = ./app,./app/stubs
...
 django_settings_module = "config.settings_dev"
+
+# relax rules for stubs
+[mypy-environ.*]
+disallow_untyped_defs = False
+disallow_incomplete_defs = False
+
+[mypy-gunicorn.*]
+disallow_untyped_defs = False
+disallow_incomplete_defs = False
```
asteiner-swisstopo, line 26: Why not the more generic `ignore_errors` (per module, not global)?

asteiner-swisstopo (commented):
Looks already pretty good, found some more opportunities to rely more on the default settings on strict mode.

mypy.ini
msom: Nice, didn't know about this!

mypy.ini
msom: The `implicit_reexport`? It's not (the other also not)

mypy.ini
msom: Not sure if this is a good idea.

mypy.ini
msom: The reason for allowing implicit reexports is to be able to directly use the import from another file, as done by you with `TranslationsSchema` in the `provider.api`. I don't see a value of forcing a user to not reuse this import...

asteiner-swisstopo: Oh my bad, you are right: I mixed it up with `no_implicit_reexport`. :hand_over_mouth:

Fine with me to keep it as it is then. I don't see why we should be forced to list exported symbols in the `__all__` attribute. Didn't even know that exists tbh.

msom: Or do something like `from xxx import yyy as yyy`, which is a bit pointless

mypy.ini
asteiner-swisstopo: I agree it looks a bit overly general. Then again: What is the advantage of having mypy report errors from an untyped package?

msom: You want some (minimal) checking of the stubs.

asteiner-swisstopo: Sure but for `gunicorn`, there are no stubs if I understand correctly. Only for `django` thanks to `django-stubs`. So maybe only set `ignore_errors` for `gunicorn`?

msom: https://github.com/geoadmin/service-control/tree/develop/app/stubs/gunicorn/app

asteiner-swisstopo (approved):

PR #29, opened 2024-10-10
PB-1044 Add interface for cognito

msom:
Added:
- a low level service class `cognito.utils.Service` for managing cognito users
- three functions in `cognito.utils.user` to be used in the API when creating, deleting and updating a user 
- a CLI command `cognito_sync` for synchronizing users (same base classes used as in #23, taken from `service-stac`)
- a new docker container `cognito-local` for local testing and development

### DIFF
--- a/app/cognito/utils/client.py
+++ b/app/cognito/utils/client.py
@@ -25,7 +25,7 @@ class Client:
     def __init__(self) -> None:
         self.endpoint_url = settings.COGNITO_ENDPOINT_URL
         self.user_pool_id = settings.COGNITO_POOL_ID
-        self.managed_flag_name = settings.COGNITO_FLAG_NAME
+        self.managed_flag_name = settings.COGNITO_MANAGED_FLAG_NAME
         self.client = client("cognito-idp", endpoint_url=self.endpoint_url)
 
     def list_users(self) -> list['UserTypeTypeDef']:
--- a/app/cognito/utils/client.py
+++ b/app/cognito/utils/client.py
@@ -19,7 +19,7 @@ def user_attributes_to_dict(attributes: list['AttributeTypeTypeDef']) -> dict[st
 class Client:
     """ A low level client for managing cognito users.
 
-    Only manages users which have managed flag defined.
+    Only manages users which have the managed flag defined.
     """
 
     def __init__(self) -> None:
--- a/app/cognito/utils/client.py
+++ b/app/cognito/utils/client.py
@@ -1,0 +1,53 @@
+from boto3 import client
+from mypy_boto3_cognito_idp.type_defs import AdminGetUserResponseTypeDef
+from mypy_boto3_cognito_idp.type_defs import UserTypeTypeDef
+
+from django.conf import settings
+
+
+class Client:
+    """ A low level client for managing cognito users. """
+
+    def __init__(self) -> None:
+        self.endpoint_url = settings.COGNITO_ENDPOINT_URL
+        self.user_pool_id = settings.COGNITO_POOL_ID
+        self.client = client("cognito-idp", endpoint_url=self.endpoint_url)
+
+    def get_users(self) -> list[UserTypeTypeDef]:
+        response = self.client.list_users(UserPoolId=self.user_pool_id)
+        return response['Users']
+
+    def get_user(self, username: str) -> AdminGetUserResponseTypeDef | None:
+        try:
+            response = self.client.admin_get_user(UserPoolId=self.user_pool_id, Username=username)
+        except self.client.exceptions.UserNotFoundException:
+            return None
+
+        return response
+
+    def create_user(self, username: str, email: str) -> UserTypeTypeDef:
+        response = self.client.admin_create_user(
+            UserPoolId=self.user_pool_id,
+            Username=username,
+            UserAttributes=[
+                {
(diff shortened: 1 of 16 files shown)


app/utils/command.py
```diff
@@ -0,0 +1,81 @@
+import inspect
+import logging
+from typing import Any
+
+from django.core.management.base import BaseCommand
+from django.core.management.base import CommandParser
+
+
+class CustomBaseCommand(BaseCommand):
```
schtibe, line 9: This is taken from service-stac, correct?

app/cognito/utils/user.py
```diff
@@ -0,0 +1,58 @@
...
+
+    client = Client()
+    username = str(user.id)
+    existing = client.get_user(username)
+    if existing is not None:
+        logger.warning("User %s already exists, updating", username)
+        client.update_user(username, user.email)
+    else:
+        client.create_user(username, user.email)
+    logger.info("User %s created", username)
```
schtibe, line 28: This will print "created" regardless of if the user was updated or created. Is that intended?

app/cognito/utils/user.py
```diff
@@ -0,0 +1,58 @@
...
+    """ Delete the given user from cognito. """
+
+    client = Client()
+    username = str(user.id)
+    existing = client.get_user(username)
+    if existing is not None:
+        client.delete_user(username)
+    else:
+        logger.warning("User %s does not exist, ignoring", username)
+    logger.info("User %s deleted", username)
```
schtibe, line 41: Same here, this will print "deleted" even if it can't because the user didn't exist. Of course, the previous message would say that, but it's in another log level, which might get confusing

app/cognito/utils/client.py
```diff
@@ -0,0 +1,53 @@
...
+            response = self.client.admin_get_user(UserPoolId=self.user_pool_id, Username=username)
+        except self.client.exceptions.UserNotFoundException:
+            return None
+
+        return response
+
+    def create_user(self, username: str, email: str) -> UserTypeTypeDef:
+        response = self.client.admin_create_user(
+            UserPoolId=self.user_pool_id,
+            Username=username,
```
schtibe, line 31: Am I right that the username is the field that identifies the records uniquely? cc @asteiner-swisstopo

schtibe: I just saw that for the moment you chose `User.id` for the username. But I think we should have a designated field for that in the model.

schtibe (commented):
Very nice!! 🚀
I have a few remarks in the code.

Some thought I just had:
Would it be good to have an attribute in the pool that is set when the user is managed from service control? So that when we clear and remove users, we only take the ones that have this attribute/flag.
Like this it would be possible to have manually created users that won't be removed by the sync command. This could prove to be handy for debugging purposes, or when we need an admin user of some sort.

app/utils/command.py
msom: Yes. It's also the same thing introduced in #23.

app/cognito/utils/user.py
msom: Yes, that was intended, but not sure if it's a good idea. Can change for more clarity.

app/cognito/utils/user.py
msom: See above.

app/cognito/utils/client.py
msom: Agree, `user.id` is a bad choice, since it's implicit and automatically set. I would also prefer a designated field.

app/utils/command.py
schtibe: :+1:

app/utils/command.py
```diff
@@ -0,0 +1,81 @@
...
+        The actual logic of the command. Subclasses must implement
+        this method.
+        """
+        raise NotImplementedError("subclasses of CustomBaseCommand must provide a handle() method")
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument('--logger', action='store_true', help='use logger configuration')
+
+
+class CommandHandler():
```
boecklic, line 22: Maybe add a comment to where this comes from (service-stac?) so that we know that changing stuff here should probably also go into service-stac

msom: Comment added.

msom:
As discussed, added a `managed` flag.

asteiner-swisstopo (approved):
Looks good to me :+1:

app/config/settings_base.py
```diff
@@ -129,3 +130,8 @@
 # https://docs.djangoproject.com/en/5.0/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
+
+# Cognito
+COGNITO_ENDPOINT_URL = env.str('COGNITO_ENDPOINT_URL', 'http://localhost:9229')
+COGNITO_POOL_ID = env.str('COGNITO_POOL_ID', 'local')
+COGNITO_FLAG_NAME = env.str('COGNITO_FLAG_NAME', 'dev:custom:managed_by_service')
```
schtibe, line 137: I think we should rename this to something `COGNITO_MANAGED_FLAG_NAME` as there might be other flags in the future!

PR #30, opened 2024-10-10
PB-1045 Add user model and GET endpoints

asteiner-swisstopo:
This adds the GET endpoints for users in service-control. The POST endpoint is done in a subsequent PR in order to unblock the works on adding the users to Cognito (see [PB-1044](https://jira.swisstopo.ch/browse/PB-1044)).

Summary of changes:
1. New app "access" as defined in [the API specs](https://2zvidp4qyh.apidog.io/api-9156067).
2. New Django model User with
   ```py
    username: CharField
    first_name: CharField
    last_name: CharField
    email: EmailField
    provider: ForeignKey
    ```
    :information_source: The fields `created` and `updated` that are listed [in the API specs](https://2zvidp4qyh.apidog.io/api-9156067) are left out for the moment.
    :information_source: We defined our own user model instead of taking [the one from `django.contrib.auth`](https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#user-model). This is to keep a minimal model and avoid confusion that Django authentication has anything to say here.
3. New endpoints:
    1. `users/{user_id}`: Returns a specific user.
        :information_source: For the response schema, I considered setting [type `EmailStr`](https://docs.pydantic.dev/2.3/usage/types/string_types/#emailstr) for field `UserSchema.email`. I didn't do so because that would introduce an additional dependency: [email-validator](https://github.com/JoshData/python-email-validator). I thought that the validation in the Django model is enough but happy to adjust if the response should also be validated.
        :information_source: Also for the response schema `UserSchema`, I considered inheriting from [ninja.ModelSchema](https://django-ninja.dev/guides/response/django-pydantic/). This would have allowed me to avoid repeating all the fields and using `UserSchema(model).from_orm()` to convert the model to the schema. I couldn't manage to get this working because Django somehow creates implicit fields like `provider_id_id` (note the double `_id` appendix). Instead, I added a function `user_to_response` similar to the other models. Please let me know if you see a way how to define the model/schema fields such that we can make use of `ninja.ModelSchema`.
    2. `users`: Returns all users.

Example response for `users/123`:

```json
{
    "username": "dude",
    "first_name": "Jeffrey",
    "last_name": "Lebowski",
    "email": "dude@bowling.com",
    "provider_id": 42,
}
```

Example response for  `users`:

```json
{
    "items": [
        {
            "username": "dude",
            "first_name": "Jeffrey",
            "last_name": "Lebowski",
            "email": "dude@bowling.com",
            "provider_id": 42,
        },
        {
            "username": "veteran",
            "first_name": "Walter",
            "last_name": "Sobchak",
            "email": "veteran@bowling.com",
            "provider_id": 42,
        },
    ]
}
```


app/config/api.py
```diff
@@ -15,3 +16,4 @@ def checker(request: HttpRequest) -> dict[str, bool | str]:
 
 api.add_router("/providers", provider_router)
 api.add_router("", distributions_router)
+api.add_router("", access_router)
```
schtibe, line 19: The patterns seem to differ here... maybe unifify the way we build these APIs?

app/access/api.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@router.get("users", response={200: UserListSchema}, exclude_none=True)
+def users(request: HttpRequest) -> UserListSchema:
+    """
+    Get all users.
+    """
+    models = User.objects.order_by("id").all()
+    schemas = [user_to_response(model) for model in models]
+    return UserListSchema(items=schemas)
```
schtibe, line 44: I guess here too we could return the data directly without going through calling UserListSchema

app/access/tests/test_api.py
```diff
@@ -0,0 +1,110 @@
...
+from access.schemas import UserSchema
+from ninja.testing import TestClient
+from provider.models import Provider
+
+from django.test import TestCase
+
+
+class ApiTestCase(TestCase):
+
+    def setUp(self):
```
schtibe, line 13: I don't know if my setup is wrong here but mypy says that the functions lack a return type definition

schtibe (approved):
Nice, just little remarks.
Now go bowling!

app/access/models.py
```diff
@@ -0,0 +1,16 @@
...
+class User(models.Model):
+
+    _context = "User model"
+
+    def __str__(self) -> str:
+        return str(self.first_name) + str(self.last_name)
+
+    first_name = models.CharField(_(_context, "First name"))
+    last_name = models.CharField(_(_context, "Last name"))
+    email = models.EmailField(_(_context, "Email"))
```
schtibe, line 14: Seeing from the cognito implementation of @msom I think you need `username` here too. Afaik the username will also be the unique identifier in the entire system? Though I'm not quite sure anymore if or what we agreed here on.

app/access/tests/test_api.py
msom: Tests should be excluded: https://github.com/geoadmin/service-control/blob/feat-PB-1045-add-user-model-and-endpoint/mypy.ini#L7. Maybe reload vscode?

app/access/models.py
msom: I would also use a designated field, independent of the email. Nonetheless, I think email needs to be unique.

app/access/tests/test_api.py
schtibe: Hmm nope... do you guys also use ms-python.mypy-type-checker in vscode?

app/config/api.py
asteiner-swisstopo: Good point, I'll set the `provider_router` also to `""` in order to align with the others.

app/access/api.py
asteiner-swisstopo: Ah yes, same as last time. Thanks, will fix :+1:

app/access/tests/test_api.py
asteiner-swisstopo: I haven't used the [Mypy Type Checker extension](https://github.com/microsoft/vscode-mypy) yet but good idea.

I had to adjust some settings such that it takes our mypy with the corresponding settings:
```json
"mypy-type-checker.args": [
    "[\"--config-file=mypy.ini\"]"
],
"mypy-type-checker.importStrategy": "fromEnvironment"
```
Would it help to mention this in the README or is this too IDE specific?

app/access/models.py
asteiner-swisstopo: Hm ok, I was not aware of that there is already [a PR with for the interface with cognito](https://github.com/geoadmin/service-control/pull/29).

Having a dedicated `username` sounds like a good idea for when emails change. I can add that.

I can also try to add a constraint that the email should be unique in the DB. From what I understand, setting [`unique=True`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#unique) in the constructor should be enough.

msom: The cognito PR is a draft and I will adapt it to what you'll implement.

msom (approved):
Looks good.

Regarding django vs. ninja validation: Have you tried what happens if in invalid email is provided (might be worth a test)? My guess is that relying on the django validator will ultimately result in an aborted transaction, where using the ninja validator will probably not. If true, using the ninja validator might be a bit cleaner.

app/access/models.py
msom: The username should be unique as well.

adk-swisstopo: I agree we should not use e-mail as username because we probably want to be able to change e-mail. I have no strong opinion on e-mail uniqueness. The use case @boecklic describes can easily work around that with [sub-addressing](https://en.wikipedia.org/wiki/Email_address#Sub-addressing) on their side (i.e. foo+bar@baz.quux) but I guess it's nice to let users use whatever address they control, be it duplicated or not.

However I am unclear why we need first_name and last_name. What about a [single full_name field](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)? However, given enough diverse users, deriving a username from the full name might be more tricky than it seems. Some entropy and checking against existing users will be required in any case. Would be nice to let the user pick their username even if we suggest one by default.

boecklic: I find it useful to be able to clearly distinguish between first_ and last_name, e.g. for ordering and be sure we have both. If it's just a fullname field we'll have a mix of 'first_name last_name', 'first_name', 'last_name first_name' etc.

asteiner-swisstopo: Thanks for your inputs.

For the moment, I change the following:
- `username`: Primary key, no defaults for simplicity. We can add that later.
- `email`: Unchanged, no requirement for uniqueness.
- `first_name`, `last_name`: Leave it as it is. I like the single `full_name` though.

I propose we quickly discuss the details in person tomorrow.

adk-swisstopo: Recommended reading before making a decision wrt {full,first,last}_name: https://www.w3.org/International/questions/qa-personal-names.en.html

app/access/tests/test_api.py
schtibe: I would actually rather add this to [geoadmin/ansible-bgdi-dev](https://github.com/geoadmin/infra-ansible-bgdi-dev/pull/165)

asteiner-swisstopo:
> Looks good.
> 
> Regarding django vs. ninja validation: Have you tried what happens if in invalid email is provided (might be worth a test)? My guess is that relying on the django validator will ultimately result in an aborted transaction, where using the ninja validator will probably not. If true, using the ninja validator might be a bit cleaner.

As discussed in chat, [Django's EmailField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#emailfield) only affects for validation. Not model validation. So no exception is raised when saving a User with an invalid email. I added a corresponding test to demonstrate that.

For the moment, we agreed on taking the extra dependency and add `email-validator` such that we can use [pydantics EmailStr](https://docs.pydantic.dev/2.3/usage/types/string_types/#emailstr). To be discussed in the team whether we want to validate only input/output at the endpoints or also when storing data to the DB.

app/access/models.py
ltshb: I would not add a `full_name` field in database, this one can be formed in a frontend application when needed by using `first` and `last`. `username` is usually the name that the user will need to enter when authenticating, and should be therefore memorized by the user (using a hash in username is quite uncommon). And because one single real person might have several account with different permission, using a username with first-last-hash will not be good. Therefore I would let the user define his username (we can still let first-last as default) and we need to make sure that the username is unique. 
Due to technical user name, we should also add an optional field `description`, so that the user can have multiple account and can describe them if he want (as we have technical users in github, the `description` field would be a very nice feature).

app/access/tests/test_api.py
asteiner-swisstopo: Nice, I see that you added it already [in this PR to infra-ansible-bgdi-dev](https://github.com/geoadmin/infra-ansible-bgdi-dev/pull/165). I resolve this thread, then.

app/access/models.py
adk-swisstopo: > [full_name] can be formed in a frontend application when needed by using first and last

Not in the general case, no. Might be good enough for most "Swiss" names though.

asteiner-swisstopo: To conclude the discussion: I go with `first_name` / `last_name` because
1. there might be a benefit in the future when connecting a user to a different user directory (eIAM?).
2. it is safe to assume that users are in Switzerland and have a sufficient workaround in case their name does not directly map to `first_name` / `last_name`.
3. A poll in the team showed that most are for this option.

I added the `username` as the primary key.

asteiner-swisstopo:
Hi @schtibe and @msom,

I have adapted the following after our discussion yesterday:
1. Add field `User.username` as primary key.
2. Leave `User.first_name` / `User.last_name` as it is.
3. Overwrite `User.clean()` and `User.save()` such that the email es validated every time we try to write to the database. 
4. Remove the type checks in `UserSchema.email` and `User.email`. Both are not necessary anymore with the validation in `User.clean()`. This also allows us to avoid the additional dependency `email-validator`.

Would you mind having another look please? :innocent:

app/access/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+        ('provider', '0006_alter_provider_acronym_it_alter_provider_acronym_rm_and_more'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='User',
+            fields=[
+                (
+                    'username',
+                    models.CharField(primary_key=True, serialize=False, verbose_name='User name')
```
msom, line 22: Why use a `primary_key` rather than the default `AutoField` and `unique`?

app/access/models.py
```diff
@@ -0,0 +1,37 @@
...
+
+    _context = "User model"
+
+    def __str__(self) -> str:
+        return str(self.first_name) + str(self.last_name)
+
+    username = models.CharField(_(_context, "User name"), primary_key=True)
+    first_name = models.CharField(_(_context, "First name"))
+    last_name = models.CharField(_(_context, "Last name"))
+    email = models.CharField(_(_context, "Email"))
```
msom, line 19: Why not using the [EmailField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#emailfield), this should use the `EmailValidator`, no?

app/access/migrations/0001_initial.py
asteiner-swisstopo: Not sure what you mean. Do you suggest having two fields, one a [numerical `AutoField`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#autofield) with an ID given by Django and additionally a `CharField` with `unique=True`? I don't see a reason to have both a numerical ID and a textual ID... can you explain?

app/access/models.py
asteiner-swisstopo: Because it would duplicate what is done in `Model.clean()`: [`validate_email()`](https://docs.djangoproject.com/en/5.1/ref/validators/#validate-email) is just a wrapper for `EmailValidator`.

There would be no harm in this at the moment but having it in two places risks having it out of sync at some point.

No strong opinion though, would you prefer to have it in both places?

app/access/migrations/0001_initial.py
msom: I think it is in general a bad idea to mix primary keys with information, since they will be used for foreign keys and if you want to change them (say in this case you have a typing error in the username you want fix), you need cascading. I've made very bad experience using this approach.

app/access/models.py
msom: If I understand correctly, django will call the validator of an email field itself when doing a full clean, so overwriting clean wouldn't be necessary. But not 100% sure about this.

app/access/migrations/0001_initial.py
asteiner-swisstopo: Good point, makes total sense.

Will adapt it to have two fields:
- An implicit `id` of type `BigAutoField`
- A `username` of type `CharField(unique=True)`

app/access/models.py
asteiner-swisstopo: I checked and you are right! As `Model.save()` now runs `Model.full_clean()`, it also runs the validators attached to the model fields. Thanks, will fix!

app/config/api.py
schtibe: This will probably lead to conflicts with https://github.com/geoadmin/service-control/pull/26/files#diff-f798933a06a1a689267b37dbf7ba771d328cb5619ae1aa9517288bfdc76d3161R7

schtibe (approved):

PR #31, opened 2024-10-22
PB-1045 Add POST endpoint to create User

asteiner-swisstopo:
This is a follow-up of [the PR](https://github.com/geoadmin/service-control/pull/30) that added the User model and their corresponding GET endpoints.

Summary of changes:
1. Add a POST endpoint that lets you create a `User` through the REST API.
    :information_source: This returns the created user and HTTP status codes according to [the API specs](https://2zvidp4qyh.apidog.io/api-10771349):
        - `201` (Created) if the User was created as expected
        - `404` (Not Found) if there is no provider with the given provider ID
        - `409` (Conflict) if there is already a record with the same username
        - `422` (Unprocessable Content) if there is any other invalid value
2. Helper function to check if the `ValidationError` raised by the model validation contains a violation of the unique constraint on `User.username` (marked by `ValidationError.code == 'unique'`). I find that somewhat shaky, the `ValidationError` class [is super flexible](https://github.com/django/django/blob/df6013b2b4e93ed6d127c2f572e6de0ba46d1d6a/django/core/exceptions.py#L134). Yet I would like to avoid having too many `isinstance(...)` / `hasattribute(...)`. In the unit test for the API endpoint, I observed that model validation creates the `ValidationError` always in the same way (something like `ValidationError.error_dict == {"field1": [ValidationError(...), ValidationError(...)], "field2": [...]}`. So maybe it is good enough but open to better suggestions.
3. Minor refactoring in unit tests `test_api.py` to reuse `TestClient` instead of creating it again in each test.

Example response for a successful POST `api/users` (code 201):

```json
{
    "username": "donny",
    "first_name": "Theodore Donald",
    "last_name": "Kerabatsos",
    "email": "donny@bowling.com",
    "provider_id": 123
}
```

Example response for a payload with invalid email and a username that already exists (code 409):

```json
{
    "detail": "['Enter a valid email address.', 'User with this User name already exists.']"
}
```

### DIFF
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -1,8 +1,6 @@
 from ninja import Router
 from provider.models import Provider
-from utils.exceptions import validation_error_to_http_error
 
-from django.core.exceptions import ValidationError
 from django.http import HttpRequest
 from django.shortcuts import get_object_or_404
 
@@ -60,15 +58,11 @@ def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
     """
     provider = get_object_or_404(Provider, id=user_in.provider_id)
 
-    try:
-        user_out = User.objects.create(
-            username=user_in.username,
-            first_name=user_in.first_name,
-            last_name=user_in.last_name,
-            email=user_in.email,
-            provider=provider
-        )
-    except ValidationError as error:
-        raise validation_error_to_http_error(error) from error
-
+    user_out = User.objects.create(
+        username=user_in.username,
+        first_name=user_in.first_name,
+        last_name=user_in.last_name,
+        email=user_in.email,
+        provider=provider
+    )
     return user_to_response(user_out)
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -47,12 +47,13 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
     return {"items": responses}
 
 
-@router.post("users", response=UserSchema)
+@router.post("users", response={201: UserSchema})
 def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
     """Create the given user and return it.
 
     Return HTTP status code
 
+        - 201 (Created) if the User was created as expected
         - 404 (Not Found) if there is no provider with the given provider ID
         - 409 (Conflict) if there is already a record with the same username
         - 422 (Unprocessable Content) if there is any other invalid value
@@ -69,4 +70,5 @@ def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
         )
     except ValidationError as error:
         raise validation_error_to_http_error(error) from error
+
     return user_to_response(user_out)
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -1,5 +1,8 @@
 from ninja import Router
+from provider.models import Provider
+from utils.exceptions import validation_error_to_http_error
 
+from django.core.exceptions import ValidationError
 from django.http import HttpRequest
 from django.shortcuts import get_object_or_404
 
@@ -42,3 +45,28 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
     models = User.objects.all()
     responses = [user_to_response(model) for model in models]
     return {"items": responses}
+
+
+@router.post("users", response=UserSchema)
+def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
+    """Create the given user and return it.
+
+    Return HTTP status code
+
+        - 404 (Not Found) if there is no provider with the given provider ID
+        - 409 (Conflict) if there is already a record with the same username
+        - 422 (Unprocessable Content) if there is any other invalid value
+    """
+    provider = get_object_or_404(Provider, id=user_in.provider_id)
+
+    try:
+        user_out = User.objects.create(
+            username=user_in.username,
+            first_name=user_in.first_name,
+            last_name=user_in.last_name,
+            email=user_in.email,
+            provider=provider
+        )
+    except ValidationError as error:
+        raise validation_error_to_http_error(error) from error
+    return user_to_response(user_out)
--- a/app/access/tests/test_api.py
+++ b/app/access/tests/test_api.py
@@ -21,6 +21,8 @@ class ApiTestCase(TestCase):
         }
         User.objects.create(**model_fields)
 
+        self.client = TestClient(router)
+
     def test_user_to_response_maps_fields_correctly(self):
 
         model = User.objects.last()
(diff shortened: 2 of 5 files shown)


app/access/api.py
```diff
@@ -42,3 +45,30 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
...
+    try:
+        user_out = User.objects.create(
+            username=user_in.username,
+            first_name=user_in.first_name,
+            last_name=user_in.last_name,
+            email=user_in.email,
+            provider=provider
+        )
+    except ValidationError as error:
+        raise validation_error_to_http_error(error) from error
```
schtibe, line 72: Is this necessary? If so, might there be a way to have this globally, as suggested here https://django-ninja.dev/guides/errors/#override-default-handler ?

app/access/api.py
```diff
@@ -42,3 +45,30 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
...
+    """
+    provider = get_object_or_404(Provider, id=user_in.provider_id)
+
+    try:
+        user_out = User.objects.create(
+            username=user_in.username,
+            first_name=user_in.first_name,
+            last_name=user_in.last_name,
+            email=user_in.email,
+            provider=provider
```
schtibe, line 69: I guess for now this is good enough, but I suspect that in the future, the provider will be deduced from the user doing the operation. Then we'll have another UserSchema, one for the input. 🤔 But this is just my thinking way ahead

app/access/tests/test_api.py
```diff
@@ -108,3 +106,82 @@ def test_get_users_returns_users_ordered_by_id(self):
...
+        assert response.status_code == 404
+
+    def test_post_users_returns_422_if_email_format_invalid(self):
+
+        invalid_email = "donny_at_bowling_dot_com"
+        payload = {
+            "username": "donny",
+            "first_name": "Theodore Donald",
+            "last_name": "Kerabatsos",
+            "email": invalid_email,
```
schtibe, line 147: Nice!

app/access/tests/test_api.py
```diff
@@ -108,3 +106,82 @@ def test_get_users_returns_users_ordered_by_id(self):
                 },
             ]
         }
+
+    def test_post_users_creates_new_user_in_db_and_returns_it(self):
+
+        payload = {
+            "username": "donny",
+            "first_name": "Theodore Donald",
+            "last_name": "Kerabatsos",
```
schtibe, line 115: I didn't know that this was his real name 😂

schtibe:
> Minor refactoring in unit tests test_api.py to reuse TestClient instead of creating it again in each test.

I think the TestClient is still being created again in each test, see https://docs.python.org/3/library/unittest.html

> The [setUp()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) and [tearDown()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown) methods allow you to define instructions that will be executed before and after each test method. They are covered in more detail in the section [Organizing test code](https://docs.python.org/3/library/unittest.html#organizing-tests). 

But in my opinion this is actually a good thing, each test then really has it's own, pristine state, and moving it to `setUp()` makes total sense.

msom:
I'm agreeing with @schtibe to implement the exceptions using the ninja exception handling (https://django-ninja.dev/guides/errors/).

I also suggest, that we use another error response format. The current solution is `str(list)`:

```
{
    "detail": "['Enter a valid email address.', 'User with this User name already exists.']"
}
```

I would suggest using either a proper string (`", ".join(list)`) 
```
{
    "detail": "Enter a valid email address., User with this User name already exists."
}
```
or proper list
```
{
    "detail": ["Enter a valid email address.", "User with this User name already exists."]
}
```

app/access/api.py
asteiner-swisstopo: Yes, I think it's necessary to do something with the exception because otherwise the user would just see a generic "500 Internal Server Error".

I was not aware of the possibility to define custom exception handlers for the whole NinjaAPI object - thanks! I'm going to try that, then.

app/access/api.py
asteiner-swisstopo: :thinking: Good point, thanks. Created a ticket as a reminder for it [PB-1084](https://jira.swisstopo.ch/browse/PB-1084).

app/access/tests/test_api.py
asteiner-swisstopo: Me neither :smile: Not surprising, given that he has to "shut the f*** up" all the time :sweat_smile:

app/access/tests/test_api.py
```diff
@@ -108,3 +106,82 @@ def test_get_users_returns_users_ordered_by_id(self):
...
+            "email": "donny@bowling.com",
+            "provider_id": Provider.objects.last().id,
+        }
+
+        response = self.client.post("users", json=payload)
+
+        assert response.status_code == 201
```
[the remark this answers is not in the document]
asteiner-swisstopo: Totally, will adapt - thanks!

app/access/tests/test_api.py
```diff
@@ -108,3 +106,82 @@ def test_get_users_returns_users_ordered_by_id(self):
...
+        non_existing_provider_id = Provider.objects.last().id + 1
+        payload = {
+            "username": "donny",
+            "first_name": "Theodore Donald",
+            "last_name": "Kerabatsos",
+            "email": "donny@bowling.com",
+            "provider_id": non_existing_provider_id,
+        }
+
+        response = self.client.post("users", json=payload)
```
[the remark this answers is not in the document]
asteiner-swisstopo: Yes why not, will add - thanks!

app/access/api.py
boecklic: @schtibe the provider of the user to create is not necessarily equal to the provider of the user that makes the call, e.g. if "we" or a platform admin (e.g. Mario) creates users

app/access/tests/test_api.py
```diff
@@ -108,3 +106,82 @@ def test_get_users_returns_users_ordered_by_id(self):
...
+            "first_name": "Theodore Donald",
+            "last_name": "Kerabatsos",
+            "email": invalid_email,
+            "provider_id": Provider.objects.last().id,
+        }
+
+        response = self.client.post("users", json=payload)
+
+        assert response.status_code == 422
+        assert response.data == {'detail': "['Enter a valid email address.']"}
```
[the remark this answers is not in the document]
boecklic: Is there more in the response error payload or is this it? In the spec I have currently sth like this 
```
{
    "code": 409,
    "description": "Username is already taken"
}
```
I can change that, no problem. The `code` information is now arguably redundant to the HTTP header code. Question is if it makes sense to have unique codes (`E1234`) that can possibly occur somewhere in the whole platform..? Then a `code` would make sense.

boecklic (commented):
As is in the spec: we should add idempotency key headers to be able to uniquely identify requests and act accordingly. As a good practice all requests to a REST API should be idempotent, i.e. submitting exactly the same request twice shouldn't change the state. With POST requests this is not the case by default, i.e. if a request adds sth. and you do the request twice you end up with two new resources. Adding a unique idempotency key can solve this problem. We can do that in a separate PR as it requires some thoughts about how we track this key.

app/access/api.py
schtibe: @boecklic Aight, good to know!

asteiner-swisstopo: Also a good point :sweat_smile: Maybe having this is as an optional field would address both scenarios. Put it on the agenda to discuss this in person.

app/access/tests/test_api.py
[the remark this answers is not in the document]
asteiner-swisstopo: Yes, this is it, only a `detail` field for an uncaught `ninja.errors.HttpError`. Ok, so I keep it as it is.

Unique internal error code: I have no experience with that but I could imagine that this would be useful for support. I'll add as a discussion point. Out of scope for this PR.

app/access/api.py
```diff
@@ -42,3 +45,30 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
...
+@router.post("users", response={201: UserSchema})
+def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
+    """Create the given user and return it.
+
+    Return HTTP status code
+
+        - 201 (Created) if the User was created as expected
+        - 404 (Not Found) if there is no provider with the given provider ID
+        - 409 (Conflict) if there is already a record with the same username
+        - 422 (Unprocessable Content) if there is any other invalid value
```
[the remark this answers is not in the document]
asteiner-swisstopo: I am not really sure either. I wouldn't say it makes clients' life substantially complicated but it might be work for nothing on our side to distinguish these cases.

I keep it as it is for the moment and add it as a point to discuss.

msom (approved):

PR #32, opened 2024-10-23
Add bandit security checker

msom:
I previously had good experience using [bandit](https://bandit.readthedocs.io) for security checks and would suggest to use it here as well (probably best in the pipeline).

What are your thoughts on this @boecklic @schtibe @asteiner-swisstopo @adk-swisstopo @ltshb @benschs (feel free to include more people to the discussion)

On a side note: There is no `RFC` label [as documented in the workflow](https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#pr-for-request-for-comment-rfc)

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,0 +1,2 @@
+[bandit]
+exclude = tests
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ YAPF := $(PIPENV_RUN) yapf
 ISORT := $(PIPENV_RUN) isort
 PYLINT := $(PIPENV_RUN) pylint
 MYPY := $(PIPENV_RUN) mypy
+BANDIT := $(PIPENV_RUN) bandit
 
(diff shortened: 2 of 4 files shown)


ltshb (commented):
I've nothing against it, but after a very quick check at the https://bandit.readthedocs.io/en/latest/plugins/index.html some of the issues will already be found by pylint, so I'm not sure it will add a plus.

NOTE: regarding the PR labels you can manually create them when missing, but I've created a draft here https://github.com/geoadmin/infra-terraform-github-bgdi/pull/118 to manage them via terraform

asteiner-swisstopo:
Nice initiative! Never heard of it but seems legit to me.

It is made by [PYCQA](https://pycqa.org/) who also houses the more popular pylint and flake8. So less concerned about the overlap with pylint as @schtibe.

How about showing this in the next tech talk session?

benschs (approved):

schtibe (approved):

PR #33, opened 2024-10-24
PB-1026 Add settings for tests

msom:
Introducing the BOD import command in service control requires setting a `TESTING` flag during tests. This is done by using a dedicated `test_settings.py`. To be able to run the tests in the feature branch, it is necessary to create this file and change the configuration previously.

This is related to https://github.com/geoadmin/infra-terraform-bgdi-builder/pull/300


schtibe (approved):
:+1:

PR #34, opened 2024-10-30
Add DELETE endpoint for users

msom:
Adds a DELETE endpoint.

Returns:
- 204 (No Content) if the user has been deleted.
- 404 (Not Found) if no user exists with the given username.
- 500 (Internal Server Error) if there is an inconsistency with Cognito (either user is unmanaged or does not exists). This should not occur and is therefore treated as an error.
- 503 (Service Unavailable) if Cognito cannot be reached.

A 204 response contains no response body; the 404 and 500 responses contain the default Ninja response body (for now, see PB-1114), while the 503 response contains `{"code": 503, "description": "Service Unavailable"}`.

The 503 error is implemented as a custom exception handler and will also apply to the create and update endpoints without any additional code changes.

### DIFF
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -1,10 +1,13 @@
 from cognito.utils.user import create_cognito_user
+from cognito.utils.user import delete_cognito_user
 from ninja import Router
 from ninja.errors import HttpError
 from provider.models import Provider
 
 from django.db import transaction
+from django.http import Http404
 from django.http import HttpRequest
+from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 
 from .models import User
@@ -75,3 +78,26 @@ def create(request: HttpRequest, user_in: UserSchema) -> UserSchema:
         if not created:
             raise HttpError(500, "Internal Server Error")
     return user_to_response(user_out)
+
+
+@router.delete("users/{username}")
+def delete(request: HttpRequest, username: str) -> HttpResponse:
+    """
+    Delete the user with the given username.
+
+    Return HTTP status code
+    - 204 (No Content) if the user has been deleted
+    - 404 (Not Found) if there is no user with the given username
+    - 500 (Internal Server Error) if there is inconsistency with cognito
+    - 503 (Service Unavailable) if cognito cannot be reached
+    """
+
+    with transaction.atomic():
+        user_to_delete = User.objects.select_for_update().filter(username=username).first()
+        if not user_to_delete:
+            raise Http404("Not Found")
+        deleted = delete_cognito_user(user_to_delete)
+        if not deleted:
+            raise HttpError(500, "Internal Server Error")
+        user_to_delete.delete()
+        return HttpResponse(status=204)
--- a/app/access/tests/test_api.py
+++ b/app/access/tests/test_api.py
@@ -249,3 +249,47 @@ class ApiTestCase(TestCase):
         assert response.data == {'code': 503, 'description': 'Service Unavailable'}
         assert User.objects.count() == 1
         assert create_cognito_user.called
+
+    @patch('access.api.delete_cognito_user')
+    def test_delete_user_deletes_user(self, delete_cognito_user):
+        delete_cognito_user.return_value = True
+
+        response = self.client.delete("users/dude")
+
+        assert response.status_code == 204
+        assert response.content == b''
+        assert User.objects.count() == 0
+        assert delete_cognito_user.called
+
+    @patch('access.api.delete_cognito_user')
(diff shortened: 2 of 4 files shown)


app/access/api.py
```diff
@@ -42,3 +45,22 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
...
+    """
+    Delete the user with the given username.
+
+    Return HTTP status code
+    - 204 (No Content) if the user has been deleted
+    - 404 (Not Found) if there is no user with the given username
+    - 500 (Internal Server Error) if there is inconsistency with cognito
+    - 503 (Service Unavailable) if cognito cannot be reached
+    """
+    model = get_object_or_404(User, username=username)
```
boecklic, line 61: I'd name this `user` or `object` since we don't delete the model itself, rather an object or instance of a model

app/access/api.py
```diff
@@ -42,3 +45,22 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
...
+    deleted = delete_user(model)
+    if not deleted:
+        raise HttpError(500, "Internal Server Error")
+    model.delete()
```
boecklic, line 65: I think we should make this whole part atomic and put them in a transaction acquiring a row-lock on the respective object (`select_for_update`). 
This would avoid race conditions and allow to better distinguish different use-cases (e.g. difference between service and cognito vs. user cannot be deleted for other reasons.) This is specially important for logging: currently you use `warning` level when the user cannot be deleted from cognito (https://github.com/geoadmin/service-control/blob/develop/app/cognito/utils/user.py#L36), depeding on the case this is rather `critical` situation in my opinion since it shouldn't happen.

boecklic (changes requested):

app/access/api.py
msom: It is named `model` for consistency with the other API functions in this file. I don't mind renaming this, but then we also need to:
- rename `model`-variables in the other functions for consistency (probably best to do this in the other APIs as well)
- rename the API endpoint functions (since the get endpoint is also called `user`; probably best to do this in the other APIs as well)

app/access/api.py
msom: Good point! I adjusted it using the transaction (and also changing log level to critical)

app/access/tests/test_api.py
```diff
@@ -204,3 +209,43 @@ def test_post_users_returns_409_and_reports_all_errors_if_multiple_things_amiss(
         assert response.data == {
             'detail': ["Enter a valid email address.", "User with this User name already exists."]
         }
+
+    @patch('access.api.delete_user')
+    def test_delete_user_deletes_user(self, delete_user):
+        delete_user.return_value = True
```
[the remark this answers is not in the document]
msom: I like this way better, and it is the same as in the other patched test.

app/config/api.py
```diff
@@ -61,6 +62,19 @@ def exception_to_response(request: HttpRequest, exception: Exception) -> HttpRes
...
 
 
+@api.exception_handler(EndpointConnectionError)
+def handle_cognito_connection_error(
+    request: HttpRequest, exception: EndpointConnectionError
+) -> HttpResponse:
+    return api.create_response(
+        request,
+        {
+            "code": 503, "description": "Service Unavailable"
```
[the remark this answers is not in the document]
msom: Let's continue this discussion in PB-1114 (feat-pb-1114-default-error-handlers).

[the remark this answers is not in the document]
msom: See comment above.

app/access/api.py
```diff
@@ -66,3 +71,26 @@ def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
...
+    """
+
+    with transaction.atomic():
+        model = User.objects.select_for_update().filter(username=username).first()
+        if not model:
+            raise Http404("Not Found")
+        deleted = delete_user(model)
+        if not deleted:
+            raise HttpError(500, "Internal Server Error")
+        model.delete()
```
schtibe, line 95: Did we not want to deactivate a user on delete?

msom: Yes, in a second step. There is currently none of the required fields available. If I remember correctly, this is a separate ticket.

asteiner-swisstopo (approved):

app/access/api.py
msom: Renamed

app/access/api.py
msom: As discussed in person: this is covered in PB-1070

boecklic (approved):
:+1:

schtibe (approved):

PR #35, opened 2024-10-30
Use custom exception handlers

msom:
Overwrite the default ninja error handlers with custom error handlers.

### DIFF
--- a/app/config/api.py
+++ b/app/config/api.py
@@ -1,14 +1,21 @@
+from logging import getLogger
+
 from access.api import router as access_router
 from distributions.api import router as distributions_router
 from ninja import NinjaAPI
+from ninja.errors import AuthenticationError
+from ninja.errors import HttpError
+from ninja.errors import ValidationError as NinjaValidationError
 from provider.api import router as provider_router
 from utils.exceptions import contains_error_code
 from utils.exceptions import extract_error_messages
 
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError as DjangoValidationError
 from django.http import Http404
 from django.http import HttpRequest
 from django.http import HttpResponse
+
+logger = getLogger(__name__)
 
 api = NinjaAPI()
 
@@ -17,8 +24,10 @@ api.add_router("", distributions_router)
 api.add_router("", access_router)
 
 
-@api.exception_handler(ValidationError)
-def validation_error_to_response(request: HttpRequest, exception: ValidationError) -> HttpResponse:
+@api.exception_handler(DjangoValidationError)
+def handle_django_validation_error(
+    request: HttpRequest, exception: DjangoValidationError
+) -> HttpResponse:
     """Convert the given validation error  to a response with corresponding status."""
     error_code_unique_constraint_violated = "unique"
     if contains_error_code(exception, error_code_unique_constraint_violated):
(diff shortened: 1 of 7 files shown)


app/config/api.py
```diff
@@ -29,35 +38,75 @@ def validation_error_to_response(request: HttpRequest, exception: ValidationErro
...
-
-    Without this, we would only see "Not found" in the response even if the
-    original exception message is more detailed like "No User matches the given query.".
-    """
+def handle_404_not_found(request: HttpRequest, exception: Http404) -> HttpResponse:
     return api.create_response(
         request,
-        {"detail": str(exception)},
+        {
+            "code": 404, "description": "Resource not found"
```
asteiner-swisstopo, line 53: Do we need the `"code"` if we already have it in the `status`?

asteiner-swisstopo (commented):
Very cool with the custom TestClient. The only doubt I have is with the redundant "code" in the response content.

app/config/api.py
msom: I've taken this from the API (draft) specs. I general, I think it is a good idea to also include the status code in the response body, as it can make my live as a (programming) user of this API easier.

app/config/api.py
```diff
@@ -29,35 +38,75 @@ def validation_error_to_response(request: HttpRequest, exception: ValidationErro
...
+def handle_ninja_validation_error(
+    request: HttpRequest, exception: NinjaValidationError
+) -> HttpResponse:
+    messages: list[str] = []
+    for error in exception.errors:
+        messages.extend(error.values())
+    return api.create_response(
+        request,
+        {
+            "code": 422, "description": messages
```
schtibe, line 104: Apparently 422 is for WebDAV. Maybe 400 would be better here, it's also more widely used. But there's a whole range of dicussions around that [here](https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data)

app/config/tests/test_api.py
```diff
@@ -0,0 +1,102 @@
...
+
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.http import Http404
+from django.http import HttpRequest
+from django.test import TestCase
+
+
+class ApiTestCase(TestCase):
+
+    def test_checker_url(self):
```
schtibe, line 16: I think this one is semantically wrong here. The checker isn't part of the API.

app/utils/testing.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    def __init__(
+        self,
+        router_or_app: NinjaAPI | Router,
+        headers: dict[str, str] | None = None,
+    ) -> None:
+        super().__init__(router_or_app=router_or_app, headers=headers)
+
+        # In order to be able to test exception handling, we need to attach the
```
schtibe, line 31: I'm wondering if it would've been easier to stop using the NinjaAPI test client and instead use the more general django one. Then you have the full call stack and test the request as it really would be coming from external.

app/config/api.py
asteiner-swisstopo: I don't see the advantage but, if that is common practice, fine with keeping it.

asteiner-swisstopo (approved):

app/config/tests/test_api.py
msom: Will split to two files, one for the checker and one for the error handlers.

app/utils/testing.py
msom: Maybe add this to the backlog of improvements as it is currently working like this?

app/config/api.py
msom: Let's discuss and make a decision today.

app/config/api.py
msom: Let's discuss and make a decision today.

msom: As discussed today: We'll use the same response structure as with [STAC](https://data.geo.admin.ch/api/stac/static/spec/v0.9/api.html#tag/Data/operation/getCollections): code and description

app/config/api.py
msom: As discussed in person: keep 422 for now.

app/utils/testing.py
msom: As discussed in person: Add this to the stack of technical dept.

PR #36, opened 2024-10-30
PB-1090 Create user also in cognito

msom:
Add connection from POST to /users to cognito

### DIFF
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -1,6 +1,9 @@
+from cognito.utils.user import create_cognito_user
 from ninja import Router
+from ninja.errors import HttpError
 from provider.models import Provider
 
+from django.db import transaction
 from django.http import HttpRequest
 from django.shortcuts import get_object_or_404
 
@@ -46,7 +49,7 @@ def users(request: HttpRequest) -> dict[str, list[UserSchema]]:
 
 
 @router.post("users", response={201: UserSchema})
-def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
+def create(request: HttpRequest, user_in: UserSchema) -> UserSchema:
     """Create the given user and return it.
 
     Return HTTP status code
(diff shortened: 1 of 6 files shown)


app/config/api.py
```diff
@@ -58,6 +59,19 @@ def exception_to_response(request: HttpRequest, exception: Exception) -> HttpRes
...
 
 
+@api.exception_handler(EndpointConnectionError)
+def handle_cognito_connection_error(
+    request: HttpRequest, exception: EndpointConnectionError
+) -> HttpResponse:
+    return api.create_response(
+        request,
+        {
+            "code": 503, "description": "Service Unavailable"
```
asteiner-swisstopo, line 69: Same as in #34 : Might want to [omit "code" and rename "description" to "detail"](https://github.com/geoadmin/service-control/pull/34#discussion_r1822801880).

asteiner-swisstopo (commented):
Looks good, basically the same as in #34 :+1:  Also left the same comments as there :sweat_smile:

So probably worth bringing #34 to an end before moving on with the others.

app/access/tests/test_api.py
```diff
@@ -140,7 +147,9 @@ def test_post_users_creates_new_user_in_db_and_returns_it(self):
         assert response.status_code == 201
         assert response.data == payload
 
-    def test_post_users_returns_404_if_provider_id_does_not_exist(self):
+    @patch('access.api.create_user')
+    def test_post_users_returns_404_if_provider_id_does_not_exist(self, create_user):
+        create_user.return_value = False
```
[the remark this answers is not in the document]
msom: See #34

app/config/api.py
msom: See #35

app/access/api.py
```diff
@@ -1,6 +1,9 @@
+from cognito.utils.user import create_user
```
schtibe, line 1: Maybe rename this function to `create_cognito_user` so it's more clear in the code below. See comment there.

app/access/api.py
```diff
@@ -55,14 +58,20 @@ def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
...
-    )
+    with transaction.atomic():
+        user_out = User.objects.create(
+            username=user_in.username,
+            first_name=user_in.first_name,
+            last_name=user_in.last_name,
+            email=user_in.email,
+            provider=provider
+        )
+        created = create_user(user_out)
```
schtibe, line 74: Semantically, you create a user then you create a user :D Maybe rename the function, see comment above.

app/access/api.py
```diff
@@ -55,14 +58,20 @@ def create_user(request: HttpRequest, user_in: UserSchema) -> UserSchema:
...
+        if not created:
+            raise HttpError(500, "Internal Server Error")
```
schtibe, line 76: I think we should log this, ideally with a reference to the call (like adding the username in the log message) so we can trace it down.

schtibe: Nevermind, just realized that the utils.creat_user does that

schtibe (approved):

asteiner-swisstopo (approved):

app/access/api.py
msom: Technically, you create two users: One in the DB and one in cognito. But sure, why not.

schtibe: Yeah I mean if you trace back to the import what `create_user` is then it's clear, just from only seeing the code here it's a bit confusing :grin:

PR #37, opened 2024-11-05
PB-1109: Add PUT endpoint to update user

asteiner-swisstopo:
Adds PUT endpoint `users/{username}` to the the REST API of service-control that allows you to update a user.

Summary of changes:

1. Add new PUT endpoint `users/{username}` that returns HTTP status code
    - 200 (OK) if the user has been updated successfully
    - 404 (Not Found) if there is no user with the given username
    - 500 (Internal Server Error) if there is an inconsistency with Cognito or if there is no provider with the given ID
    :warning: This is is the only difference in behaviour to the DELETE in #34
    - 503 (Service Unavailable) if Cognito cannot be reached

    :information_source: This uses PUT and not PATCH because we deem it good enough to start with and it's supposedly easier to implement. It's a bit annoying maybe that the user has to specify all fields, even those she/he does not want to change. But then again they can those values via the GET.

2. Rename `cognito.utils.user.update_user` to `update_cognito_user` to put more clearly indicate that this is about Cognito and not about the User model in Django. See the discussion in #34.

Example payload to PUT `users/{username}`:

```json
{
    "username": "dude",
    "first_name": "Jeff",
    "last_name": "Bridges",
    "email": "tron@hollywood.com",
    "provider_id": 123,
}
```

Example response if provider does not exist:

```json
{
    "code": 500,
    "description": "Provider does not exist"
}
```

### DIFF
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -125,8 +125,8 @@ def update_user(
 
     Return HTTP status code
     - 200 (OK) if the user has been updated successfully
-    - 404 (Not Found) if there is no user with the given username or if there is
-      no provider with the given ID
+    - 400 (Bad Request) if there is no provider with the given ID
+    - 404 (Not Found) if there is no user with the given username
     - 500 (Internal Server Error) if there is an inconsistency with Cognito
     - 503 (Service Unavailable) if Cognito cannot be reached
     """
@@ -136,7 +136,7 @@ def update_user(
             raise Http404()
 
         if not Provider.objects.filter(id=user_in.provider_id).exists():
-            raise HttpError(HTTPStatus.NOT_FOUND, "Provider does not exist")
+            raise HttpError(HTTPStatus.BAD_REQUEST, "Provider does not exist")
 
         for attr, value in user_in.dict(exclude_unset=True).items():
             setattr(user_object, attr, value)
--- a/app/access/api.py
+++ b/app/access/api.py
@@ -118,8 +118,10 @@ def delete(request: HttpRequest, username: str) -> HttpResponse:
 
 
 @router.put("users/{username}")
-def update_user(request: HttpRequest, username: str, user_in: UserSchema) -> HttpResponse:
-    """Update the given user with the given user data.
+def update_user(
+    request: HttpRequest, username: str, user_in: UserSchema
+) -> HttpResponse | UserSchema:
+    """Update the given user with the given user data and return it.
 
     Return HTTP status code
     - 200 (OK) if the user has been updated successfully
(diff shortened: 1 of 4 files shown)


app/cognito/utils/user.py
```diff
@@ -37,7 +37,7 @@ def delete_user(user: User) -> bool:
     return deleted
 
 
-def update_user(user: User) -> bool:
```
msom, line 40: We should also change the log from `warning` to `critical`.

app/access/api.py
```diff
@@ -75,3 +80,32 @@ def create(request: HttpRequest, user_in: UserSchema) -> UserSchema:
...
+    - 500 (Internal Server Error) if there no provider with the given ID
+    - 503 (Service Unavailable) if Cognito cannot be reached
+    """
+    with transaction.atomic():
+        user_object = User.objects.select_for_update().filter(username=username).first()
+        if not user_object:
+            raise Http404()
+
+        if not Provider.objects.filter(id=user_in.provider_id).exists():
+            raise HttpError(HTTPStatus.INTERNAL_SERVER_ERROR, "Provider does not exist")
```
msom, line 102: I would use some sort of 4xx response since this is under the control of the user of the API.

app/cognito/utils/user.py
asteiner-swisstopo: Right, will fix - thanks!

app/access/api.py
asteiner-swisstopo: Good point, will change to 404 :wrench:

msom (approved):

app/access/api.py
schtibe: Sorry for nitpicky, I wonder if this should be 403 Bad Request, as 404 would point to the resource that is being updated itself, i.e. the user here.

schtibe (approved):
Just small comments, looking good otherwise :+1:

app/access/api.py
boecklic: I think `422` is mostly used for webdav so I'd probably use `400 Bad Request`.

app/access/api.py
```diff
@@ -101,3 +104,34 @@ def delete(request: HttpRequest, username: str) -> HttpResponse:
...
+    - 500 (Internal Server Error) if there is an inconsistency with Cognito
+    - 503 (Service Unavailable) if Cognito cannot be reached
+    """
+    with transaction.atomic():
+        user_object = User.objects.select_for_update().filter(username=username).first()
+        if not user_object:
+            raise Http404()
+
+        if not Provider.objects.filter(id=user_in.provider_id).exists():
+            raise HttpError(HTTPStatus.NOT_FOUND, "Provider does not exist")
```
boecklic, line 128: was mentioned previously: should be `400`

boecklic (approved):

app/access/api.py
asteiner-swisstopo: Ah yes, thanks for the reminder. Fixed!

PR #38, opened 2024-11-12
PB-1074: deactivate user

benschs:
`DELETE /api/users/{username}` only soft deletes the user by setting a new field `deleted_at` to the current timestamp, as well as only disabling the user in cognito. These soft deleted users are no longer returned in `GET` requests and are also no longer displayed in the admin UI.

Currently there is no way to "hard delete" users. There is still an open question of when/who could do this.


msom:
I'm not sure if it is a good idea to have no possibility to hard-delete users (not even with the console). I would rather have the possibility to delete the user in the admin UI (but not the API) and have a new [action](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/actions/) to disable them.

Slightly related: We should also synchronize the enabled state in cognito with `cognito_sync` (see for example [this draft](https://github.com/geoadmin/service-control/compare/develop...feat-pb-1070-disable-cognito-users#diff-4a0f2ab8addb6f962ff0bfb18aeeea33826ec8907f813bc6baef83a256c7c02dR75)). Here I would prefer to delete users in cognito if deleted in the DB and disable them if disabled.

benschs:
@msom Thanks for the feedback. I changed the implementation to allow deleting/disabling via the admin ui.
Currently deleting/disabling via admin UI does not affect cognito, but as you are working to change the model to include cognito changes, I decided not to do anything for that yet.

The only thing I am still not sure about is the staticmethod `objects_api()` on `User` used in the api endpoints. I feel this implementation is error prone that in the future we accidentally also return disabled users via api. Happy for any suggestions how to default only active users in api, but allow all users in the admin UI.

msom:
Nice! Yes, adding cognito to admin UI is [in progress](https://github.com/geoadmin/service-control/tree/feat-PB-1115-cognito-in-user-model).

Regarding the `objects_api`: How about managers as done in the [DahlBook example](https://docs.djangoproject.com/en/5.1/topics/db/managers/#modifying-a-manager-s-initial-queryset)? As the first manager defined in the class will be the default manager, it could look something like this:

```
class ApiUserManager(models.Manager):
    def get_queryset(self):
        return super().get_queryset().filter(deleted_at__isnull=True)

# filtered by default
class User(models.Model):
    objects = ApiUserManager()
    all_objects = models.Manager()

# all by default
class User(models.Model):
    objects = models.Manager()
    api_objects = ApiUserManager()
```

msom (approved):
Maybe use a manager instead of the static call (see comment above), but I'm fine with both.

benschs:
> Maybe use a manager instead of the static call (see comment above), but I'm fine with both.

Thanks. I much prefer using multiple managers. By default now only active users are returned.
Admin UI and cognito management command still get all users to work with.

app/access/admin.py
```diff
@@ -1,8 +1,20 @@
...
 
 
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):  # type:ignore[type-arg]
     '''Admin View for User'''
+    list_display = ('provider', 'username', 'deleted_at')
+    actions = ["make_disabled"]
+
+    @admin.action(description="Disable selected users")
+    def make_disabled(self, request: HttpRequest, queryset: models.QuerySet[User]) -> None:
```
asteiner-swisstopo, line 15: Why "make_disabled" instead of "disable"?

app/access/models.py
```diff
@@ -2,9 +2,18 @@
 
 from django.db import models
 from django.db.models.base import ModelBase
+from django.utils import timezone
 from django.utils.translation import pgettext_lazy as _
 
 
+class ActiveUserManager(models.Manager["User"]):
+    """ActiveUserManager filters out any users that have deleted_at set.
```
asteiner-swisstopo, line 10: I would avoid using variable names in comment when possible. Variable names change and then often the comments are out-of-sync. How about "ActiveUserManager filters out any users that marked as disabled." instead?

app/access/models.py
```diff
@@ -29,3 +47,7 @@ def save(
         """Validates the model before writing it to the database."""
         self.full_clean()
         super().save(force_insert, force_update, using, update_fields)
+
+    def disable(self) -> None:
+        self.deleted_at = timezone.now()
```
asteiner-swisstopo, line 52: Maybe add a note on why datetime.now() is not a good idea here. Not obvious as we saw when the tests failed.

app/access/models.py
```diff
@@ -16,9 +25,18 @@ def __str__(self) -> str:
     first_name = models.CharField(_(_context, "First name"))
     last_name = models.CharField(_(_context, "Last name"))
     email = models.EmailField(_(_context, "Email"))
+    deleted_at = models.DateTimeField(_(_context, "deleted at"), null=True, blank=True)
 
     provider = models.ForeignKey("provider.Provider", on_delete=models.CASCADE)
 
+    # By default only return active users
+    objects = ActiveUserManager()
+    all_objects = models.Manager()
```
asteiner-swisstopo, line 34: :thinking: Isn't it confusing to override a very common default behaviour like this?

If I do `User.objects` somewhere, I would expect to get access to all objects without any filter applied. If I wanted something specific, I would check if there is something dedicated for it, something like a `User.active_objects`. Then it's also clearly visible that here some filtering is happening. With `User.objects` I would be very confused not to get the full set.

I am not against the current solution, in a way it's elegant. Just saying that, if I had not been informed about this, I would probably be very confused not to get all objects when calling `User.objects`.

How do you see that? Maybe also @msom ?

asteiner-swisstopo (commented):
Some minor comments except the one about the `User.objects`. Maybe that is the only way to get it working and I am happy to accept, just mentioning that I would find it confusing to suddenly get only a subset of users.

app/access/admin.py
benschs: Not sure what I was thinking. Probably not a lot... will change it.

app/access/models.py
benschs: My thinking was that the "disable user" is really like a "delete user", so we usually don't want these users. The special case is that we need to get/recover the deleted users. I think if you are not informed about it, you might also be confused to receive deleted objects when calling `User.object`.

asteiner-swisstopo: Yeah right, also confusing to see users you assumed to be deleted. To me still a little less confusing because at least it behaves like every else: `objects` shows _all_ objects, not just a subset.

But good enough for me to keep it as it is :+1:

asteiner-swisstopo (approved):

PR #39, opened 2024-11-12
PB-1124 Add django authentication and permission check to API

msom:
This adds a simple authentication to the API using the django (session) authentication and the django model permission and therefore the same restrictions as in the admin UI.

This PR also gets rid of the ninja test client, as this skips all middleware - probably anyway not a very good idea. Since API tests run slower now, I also added pytest-xdist to be able to test in parallel locally.


msom:
No, currently there is no possibility to use JWT or basic auth, as the context of this change is to use django access management to secure service-control in a first step. Once auth is working and authorization is discussed, we'll use the same pattern as in service stac (see [PB-1124](https://jira.swisstopo.ch/browse/PB-1124)), hence replacing this (temporary) solution.

To use the service, the same authentication as with the admin UI is used (`sessionid` cookie), e.g.

```bash
curl --cookie sessionid=xyz -i http://127.0.0.1:8000/api/users
```

boecklic (approved):
Nice

PR #40, opened 2024-11-13
PB-1115 Move cognito sync to user model

msom:
This change moves the cognito synchronization calls to the user model, ensuring that users remain synchronized even when created, edited, disabled, or deleted via the admin UI or console.

Each operation (update, insert, disable, delete) is wrapped in a transaction that will roll back if any exception occurs. Any inconsistencies detected with cognito will raise a `CognitoInconsistencyError` exception, which behaves similarly to a `DatabaseError`: it triggers an HTTP 500 response. Note that there is no way to handle this differently in the admin UI, since form validation occurs before saving the object.

Update, disable, and delete operations also lock the row until the transaction completes. In cases of race conditions (for instance, when `get_object_or_404` retrieves the object, but it has been deleted before the row lock occurs), Django will raise a `DoesNotExist` exception (with base class `ObjectDoesNotExist`). This results in a 404 error in the API.

To resolve inconsistencies with cognito, the `sync_cognito` CLI command can be used at any time. This command retains the users in the database while updating them in cognito.

CC @asteiner-swisstopo @benschs @schtibe  : Feel free to add a review or comment, but don't feel obligated to do a full review.


msom:
While I agree that this might decrease DB perfomance, I don't think it would be possible as suggested. First, I don't think rolling back cognito is possible, for example when deleting: you would need to re-create the user and therefore also re-confirm the e-mail. Also, there will probably be unresolvable race conditions when first updating cognito and the encountering a DB conflict.

boecklic:
@benschs @msom tricky problem!
I basically second the perception that doing the call to cognito during an atomic transaction is not a good pattern. Although Cognito is not any random external service that is probably slow and unreliable, so we could argue that for AWS services it's acceptable. 
We'll face the principle problem of having distributed transactions more in the future (think about the policy DB that will come or service-control creating stac collections etc.), so we should come up with a generic way how we want to handle these cases. There are patterns on how to do this, they are however not trivial (mostly event-based). 
I think this is certainly something that we should keep on the radar and discuss in-depth at some point.
For the short term I see different options that would all be acceptable for me:
- leave it as it is, given that we call a very performant and relyable external service
- execute the two steps one after the other (and only the django part in an atomic transaction), do proper logging and roll back or resolve manually in case of exceptions (choosing the right order of service-control mutation and cognito mutation can probably help reducing tricky edge cases)
- first do the service-control mutation, second do the cognito mutation and if it fails, undo the service-control mutation

msom:
@benschs would you be ok with leaving it like this (for now)?

benschs (approved):
@msom Yes, we can leave it like this

PR #41, opened 2024-11-19
PB-1112: Use pytest

msom:
This PR improves the test suite, which previously contained a mix of Django and pytest tests. It

- moves all tests to `app/tests` to align with the structure of the other services
- removes all `__init__.py` files within test folders to prevent test code from being auto-discovered outside of pytest
- transfers the functionality of `utils/testing.py` to fixture(s) in conftest.py
- adapts the tests to use pytest


app/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+
+@fixture
+def user_factory(db):
+
+    def create_user_with_permissions(
+        username: str, password: str, permissions: list[tuple[str, str, str]]
+    ) -> Any:
+        user = get_user_model().objects.create_user(username=username, password=password)
+        for app_label, model, codename in permissions:
+            content_type = ContentType.objects.get(app_label=app_label, model=model)
```
schtibe, line 18: Maybe add a little comment here why this is necessary.

app/tests/access/test_access_api.py
```diff
@@ -0,0 +1,609 @@
...
+from access.api import user_to_response
+from access.models import User
+from access.schemas import UserSchema
+from botocore.exceptions import EndpointConnectionError
+from provider.models import Provider
+from pytest import fixture
+
+
+@fixture(name='access_data')
+def fixture_access_data(db):
```
schtibe, line 12: The name `access_data` confuses me a bit. I thought this data was for testing the "access", but below it seems to be about the user in the db to run the tests on

schtibe: OK so it's access data because this here is the tests for the access app? Maybe add a little docstring to the fixture describing the purpose of it

app/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+
+from pytest import fixture
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+
+
+@fixture
+def user_factory(db):
```
schtibe, line 11: The name is a bit ambiguous: it's about the user that's used to authenticate on the API, but that isn't exactly clear. It could be understood as the users in the db, the ones that we push to cognito.

app/tests/cognito/test_cognito_client.py
```diff
@@ -0,0 +1,255 @@
+from unittest.mock import call
+from unittest.mock import patch
+
+from cognito.utils.client import Client
+from cognito.utils.client import user_attributes_to_dict
+
+from django.conf import settings
+
+
+def cognito_user(username, managed, for_):
```
schtibe, line 10: I think it's good practice if functions contain a verb. Unless you turn this into a fixture of course.

app/tests/distributions/test_attribution_model.py
```diff
@@ -0,0 +1,130 @@
...
+
+def test_form_invalid_for_blank_mandatory_field(db):
+
+    class AttributionForm(ModelForm):
+
+        class Meta:
+            model = Attribution
+            fields = "__all__"
+
+    provider = Provider.objects.create()
```
schtibe, line 117: I guess you could theoretically omit the `db` fixture here and create the Model via `Provider` 😅

app/tests/distributions/test_distributions_api.py
```diff
@@ -0,0 +1,727 @@
...
+            rm=None,
+        ),
+        provider_id=Provider.objects.last().id,
+    )
+
+    assert actual == expected
+
+
+def test_get_attribution_returns_existing_attribution_with_default_language(
+    distribution_data, client, user_factory
```
schtibe, line 111: I'm wondering if we should be more concise with the names of the fixtures. `user_factory` is called "factory" but `distribution_data` is not. To me, `distribution_data` is also sort of a factory (even if we don't call it explicitly, it's sort of being called without argument). 
Maybe that's influenced from using https://pytest-factoryboy.readthedocs.io/en/stable/ in the past, idk. 
I'm not saying that we should now call them all `distribution_data_factory` per se, I just want to trigger some re-evaluating of the names. (This probably applies to most of the tests)

schtibe: OK now having seen all the tests maybe it's actually OK this way. I would consider renaming `user_factory` to make it clear that this is about the user that's connecting to the API, see my other comment.

schtibe (changes requested):
Very nice, thank you!

I have made some comments along the reviewing concerning the naming of things. It wasn't always a 100% apparent what is what, maybe this gives you a hint for improvements.

app/tests/conftest.py
msom: Done

app/tests/conftest.py
msom: Renamed to django_user_factory

app/tests/cognito/test_cognito_client.py
msom: Moved to global fixture `cognito_user_response_factory`

app/tests/distributions/test_attribution_model.py
msom: moved to fixture

app/tests/distributions/test_distributions_api.py
msom: changed all data fixtures to individual object fixtures

msom:
Your right, naming was confusing! Cleanup done:
- Changed all the data fixtures to individual fixtures returning single database objects
- Renamed `user_factory` to `django_user_factory`
- Moved the functionality of `cognito_user` to a global fixture `cognito_user_response_factory`
- Documented the global (factory) fixtures

schtibe (approved):
Nice, you made it even better! Thank you!

PR #43, opened 2024-11-26
PB-1106: Use cognito's preferred username

msom:
Introduces a new `user_id` field in `access.User`, which is an autogenerated 12-digit `nanoid`. This `user_id` serves as the Cognito username, while the existing `access.User.username` is used as the `preferred_username` in Cognito.

The database migration will assign a `user_id` to all existing users.

To update existing Cognito users, you can use `./manage.py cognito_sync`. This process will delete the old users in Cognito and re-create them using the new `user_id`.

Checklist


app/tests/utils/test_short_id.py
```diff
@@ -0,0 +1,16 @@
...
+    assert short_id_1 != short_id_2
+    assert len(short_id_1) == 12
+    assert len(short_id_2) == 12
+
+
+def test_generate_short_id_settings(settings):
+    settings.SHORT_ID_SIZE = 4
+    settings.SHORT_ID_ALPHABET = 'a'
+    short_id = generate_short_id()
+    assert generate_short_id() == 'aaaa'
```
benschs, line 16: Unused line `short_id = generate_short_id()` ?

benschs (approved):

PR #44, opened 2024-11-26
Improve the usability of the admin UI

msom:
Improve the usability of the admin UI by by adding filters and optimize displayed rows.

Providers:
- Use the acronym rather than the name in `__str__` so that provider filters are displayed more clearly
- Show `acronym_en` and `name_en` rather than only `name_en`

image

Users:
- Move `provider` to the back (previously one needed to click on the provider to edit the user)
- Add filters for `deleted_at` and `provider`

image

Attribution:
- Show `name_en` and `provider` rather than only `name_en`
- Add filter for `provider`

image

Dataset:
- Show `slug` and `provider` rather than only `slug`
- Add filter for `provider`

image

### DIFF
--- a/app/access/admin.py
+++ b/app/access/admin.py
@@ -8,8 +8,9 @@ from .models import User
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):  # type:ignore[type-arg]
     '''Admin View for User'''
-    list_display = ('provider', 'username', 'deleted_at')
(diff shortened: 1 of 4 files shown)


msom:
> Makes sense. Maybe also useful to have the `user_id` in the overview/filter of users.

Definitely, I will add this to the other PR!

PR #45, opened 2024-11-27
PB-1115: Validate all models on save

msom:
Currently, only the user model undergoes validation upon saving. This could result in scenarios where providers, attributions, and datasets are imported from the BOD but cannot be edited through the admin UI. Additionally, the slug field used in datasets is incompatible with existing datasets, as it does not support periods.

This PR introduces model validation across all models to ensure consistency and implements a custom slug field that supports periods.


benschs (approved):
Cool. :bug: (slug)

PR #49, opened 2024-12-10
Add a little command to create or update our superuser

schtibe:
The idea is to create a superuser via terraform and share it with SSM to the cluster. That way we can make sure the user is always existing and up-to-date with the parameter entry


msom (approved):
:superhero:

boecklic (approved):
Nice :ok_hand:

PR #57, opened 2024-12-18
Feat pb 1160 harvest package data

msom:
This PR adds a `stac_harvest` management command to the distributions application to sync package distributions with `service-stac`.

The command uses the STAC collection ID as the package distribution slug and relies on the `managed_by_stac` flag to identify package distributions managed by `service-stac`.

The command ensures that for each STAC collection, a corresponding package distribution exists by:
- Creating new package distributions if no package distribution matches the STAC collection.
- Updating package distributions with a matching STAC collection. This will update the dataset and force the `managed_by_stac` flag to be set.
- Removing package distributions with the `managed_by_stac` flag set but no matching STAC collection.

Assumptions and Behavior
- Dataset Matching: A package distribution is created only if a dataset exists with a slug matching the STAC collection ID. If no such dataset exists, no package distribution will be created.
- Provider Matching:
  - Each STAC collection must have exactly one provider, which should match the dataset’s provider (by its English name).
  - If a mismatch is detected, a warning is issued, but the package distribution is still created.
  - To suppress warnings for near-matching provider names (e.g., 'Federal Office for Agriculture - FOAG' vs. 'Federal Office for Agriculture FOAG'), a [similarity ratio](https://docs.python.org/3/library/difflib.html#difflib.SequenceMatcher.ratio) can be used.

Customization Options
- `--url`: Allows to specify the STAC endpoint. By default, the command uses the `service-stac` endpoint (https://data.geo.admin.ch/api/stac/v0.9).
- `--clear`: Deletes all package distributions previously marked as managed by `service-stac`.
- `--dry-run`: Runs the command without making any changes, allowing you to preview the actions.


msom:
@boecklic I'm not sure if introducing `pystac-client` is kind of an overkill here (see https://github.com/geoadmin/service-control/pull/57/commits/8ec3478024385424436d42312103ac1773febef5).

boecklic:
> @boecklic I'm not sure if introducing `pystac-client` is kind of an overkill here (see [8ec3478](https://github.com/geoadmin/service-control/commit/8ec3478024385424436d42312103ac1773febef5)).

Hmm, it probably is... but we know from customer feedbacks that this library is used by external to access our API, so it can actually be a good idea to use it ourselves as well to be able to identify potential issues. I'd leave it for the moment

PR #64, opened 2024-12-20
Put everything of the API under v1

schtibe:
I  think we decided on this at some point?


msom (approved):
I think so, yes. Makes sense.

asteiner-swisstopo (approved):

PR #66, opened 2025-01-07
Remove DOCKERFILE command that's useless in service-control

schtibe:
The lines were copied from service-stac but aren't relevant in this context, as the settings are built differently


msom:
What about [wsgi](https://github.com/geoadmin/service-control/blob/develop/app/wsgi.py#L40)?

PR #67, opened 2025-01-07
Set the manage.py to default to the general settings file

schtibe:
This referenced settings file is copied when building the docker image


app/manage.py
```diff
@@ -7,7 +7,7 @@
 def main() -> None:
     """Run administrative tasks."""
     # default to the prod settings. Can be overwritten with .env
```
msom, line 9: Remove this comment?

msom (approved):

PR #68, opened 2025-01-23
PB-1351 Log requests and errors to ECS

schtibe:


app/config/api.py
```diff
@@ -19,7 +22,117 @@
...
+
+    Returns:
+        dict: dict of extras
+    """
+    return {
+        'http': {
+            'request': {
+                'method': request.method or 'UNKNOWN',
+                'header': {
+                    k.lower(): v for k, v in request.headers.items()
```
msom, line 73: Does this log credentials?

app/config/api.py
```diff
@@ -19,7 +22,117 @@
...
+        data: Any,
+        *args: List[Any],
+        status: Optional[int] = None,
+        temporal_response: Optional[HttpResponse] = None,
+    ) -> HttpResponse:
+        response = super().create_response(
+            request, data, *args, status=status, temporal_response=temporal_response
+        )
+
+        if response.status_code >= 200 and response.status_code < 300:
```
msom, line 108: Maybe just use `>= 200 and < 400` and remove the next block?

msom (commented):
LGTM but not sure if we need to already think about preventing logging credentials (for example the values of `Authoriaztion` headers?

app/config/api.py
schtibe: Very valid question. I don't see the api gateway headers in the logs, but I'm not sure if that's intended or not. I have to do some more checks.

msom: I think ATM there is nothing yet but I don't know what the plan is here.

schtibe: So we'll see `geoadmin-authenticated` which is as I understand it a `bool` and `geoadmin-username`. I think it might even be valuable to be able to see these?

app/config/api.py
schtibe: Yes indeed. At some point in development I used a different log level for those between 300 and 400, but I think info should be alright. `info` is described as `Confirmation that things are working as expected.`, so I guess the 300s are OK there, WDYT?

msom: Absolutely

app/config/api.py
msom: You're right.

But how about generally skipping (the content of) the [Authorization Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization)? If we ever add Token or Basic Auth and forget to adapt the logging?

msom:
Also not sure about other Header Content, e.g. Set Cookie (https://security.stackexchange.com/questions/249442/storing-session-id-in-application-logs)

msom (approved):

PR #76, opened 2025-03-11
Update Django to 5.0.13

msom:
Resolves https://github.com/geoadmin/service-control/security/dependabot/14


asteiner-swisstopo (approved):
Sounds reasonable :+1:

PR #80, opened 2025-04-08
PB-1580: Extend dataset with title and description

benschs:
Add title and description in 5 languages to dataset model. As with `Attribution` the german, french and english translations are required. Also added fields to api responses.
Extend `bod_sync` command to also migrate title and description from the geocat_publish table. If a required lagnauge is missing, set value to `#Missing`.


app/bod/management/commands/bod_sync.py
```diff
@@ -271,6 +272,36 @@ def import_datasets(self) -> None:
...
+                missing = True
+            if not bod_meta.abstract_de:
+                bod_meta.abstract_de = "#Missing"
+                missing = True
+            if not bod_meta.abstract_fr:
+                bod_meta.abstract_fr = "#Missing"
+                missing = True
+            if not bod_meta.abstract_en:
+                bod_meta.abstract_en = "#Missing"
+                missing = True
```
asteiner-swisstopo, line 301: Maybe easier to read as

```py
required_fields = [
    bod_meta.bezeichnung_de,
    bod_meta.bezeichnung_fr,
    bod_meta.bezeichnung_en,
    bod_meta.abstract_de,
    bod_meta.abstract_fr,
    bod_meta.abstract_en,
]
for field in required_fields:
    if not field:
        field = "#Missing"
        missing = True
```

What do you think?

app/tests/bod/test_bod_sync_command.py
```diff
@@ -46,7 +47,25 @@ def fixture_bod_dataset(bod_contact_organisation):
...
+        bezeichnung_de="some_val",
+        bezeichnung_fr="some_val",
+        bezeichnung_it="some_val",
+        bezeichnung_rm="some_val",
+        bezeichnung_en="some_val",
+        abstract_de="some_val",
+        abstract_fr="some_val",
+        abstract_it="some_val",
+        abstract_rm="some_val",
+        abstract_en="some_val"
```
asteiner-swisstopo, line 64: Are these values compared anywhere in the test? If yes, setting them all to the same value risks not detecting if one is set to the value of another accidentally.

app/bod/management/commands/bod_sync.py
```diff
@@ -271,6 +272,36 @@ def import_datasets(self) -> None:
...
                 continue
 
+            # Get meta information title and description
+            bod_meta = BodGeocatPublish.objects.filter(fk_id_dataset=bod_dataset.id_dataset).first()
+            if not bod_meta:
+                # Skip as missing meta information
+                self.print(f"Dataset '{bod_dataset.id_dataset}' is missing geocat entry")
+                continue
+
+            # REMOVE THIS BLOCK ONCE DATA HAS BEEN FIXED
```
asteiner-swisstopo, line 282: I wasn't sure what is meant here before reading your comment in PB-1580. Maybe be more specific like "Remove this block once all datasets have data in the required fields".

asteiner-swisstopo (approved):
Looks good to me, only some maintainability-related suggestions.

asteiner-swisstopo:
> As discussed:
> 
>     * actually test the import of the new fields in bod_sync
> 
>     * update the link in the PR template diagram (https://swissgeoplatform.atlassian.net/wiki/spaces/PB/pages/16155637/Business+Entity+Model)

The link is already updated, see #81. My bad, only just seen your comment now.

app/bod/management/commands/bod_sync.py
benschs: I tried this approach, but I think I need to use something like this to update the field:
```
required_fields = [
    "bezeichnung_de",
    "bezeichnung_fr",
    "bezeichnung_en",
    "abstract_de",
    "abstract_fr",
    "abstract_en",
]
for field in required_fields:
    if not getattr(bod_meta, field):
        setattr(bod_meta, field, "#Missing")
```
But then I get an issue with the mypy type checking. So I left it as is.

app/tests/bod/test_bod_sync_command.py
benschs: Thanks. I extended the test cases and changed the values. Actually had a bug in the code..

PR #83, opened 2025-04-09
PB-1581: Extend dataset with geocat id

benschs:
Add field `geocat_id` to dataset model. Display link to geocat in admin UI. Sync geocat_id from bod `dataset.fk_geocat`.

- [x] Model changes documented in the [business entity model](https://swissgeoplatform.atlassian.net/wiki/spaces/PB/pages/16155637/Business+Entity+Model)


app/distributions/migrations/0012_dataset_geocat_id.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import migrations
+from django.db import models
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def populate_geocat(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    Dataset = apps.get_model('distributions', 'Dataset')
+    for obj in Dataset.objects.all():
+        obj.geocat_id = 'changeme ' + generate_short_id()
+        obj.save()
```
asteiner-swisstopo, line 15: For my understanding: Is that generated by Django or did you add that? I have never seen functions in these migration files.

app/bod/management/commands/bod_sync.py
```diff
@@ -272,6 +272,14 @@ def import_datasets(self) -> None:
                 )
                 continue
 
+            if legacy_id in (1485, 1486):
+                # Skip datasets ch.swisstopo.konsultation-lk10-flurnamen (1485) and
+                # ch.swisstopo.konsultation-lk10-siedlungsnamen (1486) as they have the same
+                # geocat_id as ch.swisstopo.landeskarte-farbe-10 (952)
+                # geocat_id is cb0f8401-c49a-4bdf-aff6-40a7015ba43a
```
asteiner-swisstopo, line 279: I would add more context: It would not be clear to me why these are skipped without the comments in PB-1581.

Maybe something like that?

```py
# Skip the following datasets as they are hidden in public BGDI services:
# - ch.swisstopo.konsultation-lk10-flurnamen (1485)
# - ch.swisstopo.konsultation-lk10-siedlungsnamen (1486)
# They were used for name verifications of the cantons in the context of the KONAM project and are to be deleted.
# Layer ch.swisstopo.landeskarte-farbe-10 (952) has the same geocat_id ("cb0f8401-c49a-4bdf-aff6-40a7015ba43a").
```

asteiner-swisstopo (approved):
Looks good to me. Crazy how many changes just adding one field causes :face_with_spiral_eyes:  Just left some minor comments on maintainability.

app/distributions/migrations/0012_dataset_geocat_id.py
benschs: I added this. The issue is that we want `geocat_id` to be unique and the migration does not set good default values. So I initially generated the file, but then edited it to create the column without the unique key, populate the rows with "dummy" unique values and then set the unique constraint.

PR #88, opened 2025-05-01
PB-1510: Update bod image to postgres 16

msom:
Django 5.2 requires at least [Postgres 14](https://docs.djangoproject.com/en/5.2/ref/databases/#postgresql-notes), BOD uses Postgres 16 as far as I know (right?)


benschs (approved):
Yes it's currently running version 16.7

PR #95, opened 2025-06-10
PB-1751: Allow stac_sync to create the default dataset

msom:
Add an option to `stac_sync` to allow the given default dataset if not yet existing.

Uses the same id for the attribution and provider. Uses `#Missing` similar to the BOD command for filling in missing values.

### DIFF
--- a/app/distributions/management/commands/stac_sync.py
+++ b/app/distributions/management/commands/stac_sync.py
@@ -7,8 +7,10 @@ from typing import cast
 from urllib.parse import urljoin
 
 from bs4 import BeautifulSoup
+from distributions.models import Attribution
 from distributions.models import Dataset
 from distributions.models import PackageDistribution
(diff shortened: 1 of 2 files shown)


app/distributions/management/commands/stac_sync.py
```diff
@@ -228,6 +285,11 @@ def add_arguments(self, parser: CommandParser) -> None:
             default="",
             help="Add packages with missing dataset to this dataset"
         )
+        parser.add_argument(
+            "--create-default-dataset",
+            action="store_true",
+            help="Create the default dataset if needed"
+        )
```
boecklic, line 292: I imagine we always (or mostly) want to create these default datasets, no? We could omit the option (or invert to opt-out) and by default just create the default objects that are needed.

boecklic (approved):

PR #111, opened 2025-08-22
Feat checkout uv and ruff

schtibe:
- [ ] Model changes documented in the [business entity model](https://swissgeoplatform.atlassian.net/wiki/spaces/PB/pages/16155637/Business+Entity+Model)


app/config/settings_prod.py
```diff
@@ -1,11 +1,11 @@
-from .settings_base import *  # pylint: disable=wildcard-import, unused-wildcard-import
+from .settings_base import *  # noqa: F403
```
ltshb, line 1: I don't like the `noqa: F403` is there no way to disable using some wording instead of ID, that something I like in pylint, it make much more clearer what is disabled

ltshb (commented):
I have nothing about new stuff, however if we want to keep the same toolchain all over the places, then it will requires some effort. I think we need for this good argument. But to me we could also use it only for new repository if it adds really a plus value.

schtibe:
> I have nothing about new stuff, however if we want to keep the same toolchain all over the places, then it will requires some effort. I think we need for this good argument. But to me we could also use it only for new repository if it adds really a plus value.

@ltshb @msom Yes I 100% agree that it should be the same everywhere. I wanted to try it out here as here it's less disruptive.

msom:
See also https://github.com/swissgeo/service-control/pull/28

PR #112, opened 2025-08-22
Feat format with ruff

schtibe:
- [ ] Model changes documented in the [business entity model](https://swissgeoplatform.atlassian.net/wiki/spaces/PB/pages/16155637/Business+Entity+Model)


msom:
@schtibe we decided yesterday in the planning to switch to these tools for new projects in swissgeo so maybe close this PR?

PR #140, opened 2026-03-09
GPS-539: added layers-js export functionality

boecklic:
This needs https://github.com/swissgeo/infra-terraform/pull/272


msom (approved):
🚀

PR #143, opened 2026-03-26
GPS-589: Switch geocat harvester to eCH-0271

msom:
Use the new eCH-0271 profile in the geocat harvester.

Changes:

- Nearly all tags have changed namespaces
- Some information are not present anymore
  - Contact online resources name and description including their translations (according to the standard this should still be available but it cannot be entered in geocat; I don't think it's very useful anyway)
  - Contact organization email
  - Contact hours of service
  - Contact instructions
  - Contact street name / number (this is now a single delivery point / address
- I removed some information which is meant for individuals rather than organizations
  - Position name
  - Individual name

Tested on geocat DEV and INT, hopefully this still works when this goes online on the 7.5.

:boom: Do not merge before 7.5.2026 :boom:


boecklic (approved):
👌 thanks

PR #151, opened 2026-04-21
GPS-616: Add table-role-arn argument to dataset export command

msom:
This PR updates the dataset export command:

- Adds table-role-arn argument to be able to run as cronjob
- Adds functionality to remove obsolete providers and datasets
- Renames the command from "export" to "dataset_export"
- Removes unused command line arguments
- Updates packages to resolve https://github.com/geoadmin/service-control/security/dependabot/49


ltshb (approved):
Looks good to me

PR #153, opened 2026-04-28
Use boto3 type serializer

msom:
No need to reinvent the wheel :)

See https://github.com/swissgeo/service-control/pull/121


boecklic (approved):
🥇 thanks for eliminating that hacky part!

